PR #7, opened 2024-04-24
Add to downstream ci

iainrussell:


iainrussell:
This PR is basically ready. I renamed the previous ci - 'legacy-ci' so that we are consistent across packages that our new ci is call ci.yml. I need to inspect the legacy one more closely to see where we have overlap, but I think this PR is already useful because we will be testing on our own platforms.

sandorkertesz (approved):

PR #13, opened 2024-05-03
Support multiple array formats

### LINKED ISSUE #9: Allow using multiple array formats
### Is your feature request related to a problem? Please describe.

_No response_

### Describe the solution you'd like

_No response_

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

### Organisation

_No response_

sandorkertesz:
Implements #9

I converted all the modules to be array format agnostic but still allowing numbers as input:

- extreme
- score
- solar
- stats
- thermo
- wind

All the methods (with one exception, see below) are working with the following array backends:

- numpy 
- cupy
- torch 

All the tests had to be heavily refactored.

Details
--------
The code goes like this:

```python
from earthkit.utils.array import array_namespace
...
def specific_humidity_from_vapour_pressure(e, p, eps=1e-4):
   ....
 
    xp = array_namespace(e, p)
    v = xp.asarray(p + (constants.epsilon - 1) * e)
    v[xp.asarray(p - e) < eps] = xp.nan
    return constants.epsilon * e / v
```

So the namespace is not passed as a kwarg but determined from the input arrays.

The tests now work with multiple array formats and numbers too:

```python
from earthkit.meteo.utils.testing import ARRAY_BACKENDS
...

@pytest.mark.parametrize(
    "vp, p, v_ref",
    [
        ([895.992614, 2862.662152, 10000], [700, 1000, 50], [0.008, 0.018, np.nan]),
        ([895.992614, 2862.662152, 100000], 700, [0.008, 0.0258354146, np.nan]),
        (895.992614, 700, 0.008),
        (100000, 700, np.nan),
    ],
)
@pytest.mark.parametrize("array_backend", ARRAY_BACKENDS)
def test_specific_humidity_from_vapour_pressure(vp, p, v_ref, array_backend):
    vp, p, v_ref = array_backend.asarray(vp, p, v_ref)
    p = p * 100
    q = thermo.array.specific_humidity_from_vapour_pressure(vp, p)

    assert array_backend.allclose(q, v_ref, equal_nan=True)

```

Dependencies
----------------
The namespace/backend code was moved to`earthkit-utils`. It is a required dependency.

Outstanding problem with cupy
---------------------------------
`cupy.quantile()` behaves differently than `np.quantile()` when there are nans in the input array. For this reason the test `test_quantiles_nan()` is disabled for cupy! This problem has to be further investigated!

Problems
----------
- `np.polynomial.polynomial.polyval()` is used but not part of the array API standard. No `polyval()` is available in torch.
- `torch.sign()` returns 0 for nan. According to the array API standard it should return nan. Numpy behaves correctly.
- The following  methods are used but not available in the array API standard. However, they are all available in numpy, cupy and torch. 

    - deg2rad
    - atleast_1d
    - fmax
    - fmin

- `np.percentile()` is not available in the array API standard. Torch only has `quantile()`, which is also in numpy/cupy. `quantile()` is not in the array API standard.
- `np.histogram2d()` is not available in the array API standard. Torch only has `histogramdd()`, which is also in numpy/cupy. `histogramdd()` is not in the array API standard.
-  The following methods are only available in numpy:
   - seterr

- The following methods have a different name in the array API standard than in numpy.
 
    - `fabs()` -> `abs()`
    - `power()` -> `pow()`
   
    On an array-api-compat namespace we can only call `pow()` and `abs()`!

    ```python
    import numpy as np
    import array_api_compat.numpy as xp
    a = np.ones(10)
    xp.power(a, 2) # this fails
    xp.pow(a, 2) # this works
    ```

Solutions
--------------------

Modifications
++++++++++++++++

- Replaced calls of `power` and `fabs` with `pow` and `abs`, respectively.

Patches
+++++++++++++++
The namespace returned by `array_api_compat.array_namespace()` is patched using methods from `earthkit-utils.array.compute`. See: `earthkit-utils.array` for details.  

Note:
-   `earthkit-utils.array.compute.percentile` is implemented with `quantile`.
-   `earthkit-utils.array.compute.histogram2d` is implemented with `histogramdd`.
-  `earthkit-utils.array.compute.seterr` does nothing and returns an empty dict

numpy:
    - `np.polynomial.polynomial.polyval` is added as `polyval`

cupy:
    - `cupy.polynomial.polynomial.polyval` is added as `polyval`
    - `earthkit-utils.array.compute.seterr` is added as `seterr`

torch:
    - `earthkit-utils.array.compute.polyval` is added as `polyval`
    - `earthkit-utils.array.compute.percentile` is added as `percentile`
    - `earthkit-utils.array.compute.histogram2d` is added as `histogram2d`
    - `earthkit-utils.array.compute.seterr` is added as `seterr`
    - `sign` is modified to treat nans correctly

other namespaces (not tested, maybe not needed)
    - `earthkit-utils.array.compute.polyval` is added as `polyval`
    - `earthkit-utils.array.compute.percentile` is added as `percentile`
    - `earthkit-utils.array.compute.histogram2d` is added as `histogram2d`
    - `earthkit-utils.array.compute.seterr` is added as `seterr`

With this we can call `polyval`, `percentile` and `histogram2d` on an array-namespace:

```python

def compute_wbpt(self, ept):
    xp = array_namespace(ept)
    t0 = 273.16
    x = ept / t0
    ....
    return ept - xp.exp(xp.polyval(x, a) / xp.polyval(x, b))

def sot(clim, ens, perc, eps=-1e4):
    xp = array_namespace(clim, ens, perc)
    clim = xp.asarray(clim)
    ens = xp.asarray(ens)
    ...
    qf = xp.percentile(ens, q=perc, axis=0)
```

TODO
--------
- Handle `np.seterr`
- Add JAX  support?


sandorkertesz:
Hi @corentincarton, @oiffrig, any thought on this?

src/earthkit/meteo/utils/namespace/torch.py
```diff
@@ -0,0 +1,35 @@
...
+    """Reimplement the sign function to handle NaNs.
+
+    The problem is that torch.sign returns 0 for NaNs, but the array API
+    standard requires NaNs to be propagated.
+    """
+    x = _xp.asarray(x)
+    r = _xp.sign(x, *args, **kwargs)
+    r = _xp.asarray(r)
+    r[_xp.isnan(x)] = _xp.nan
+    return r
```
HCookie: From what we discussed this seems like a good solution

src/earthkit/meteo/extreme/array/sot.py
```diff
@@ -32,24 +31,36 @@ def sot_func(qc_tail, qc, qf, eps=-1e-4, lower_bound=-10, upper_bound=10):
...
+    xp = array_namespace(qc_tail, qc, qf)
+    qc_tail = xp.asarray(qc_tail)
+    qc = xp.asarray(qc)
+    qf = xp.asarray(qf)
+
+    # TODO: check if this is necessary
     # avoid divided by zero warning
-    err = np.seterr(divide="ignore", invalid="ignore")
+    try:
+        err = xp.seterr(divide="ignore", invalid="ignore")
```
HCookie: Is this local to only numpy?
Could it make sense to check if numpy run func otherwise skip?

src/earthkit/meteo/solar/array/solar.py
```diff
@@ -70,20 +72,26 @@ def cos_solar_zenith_angle(date, latitudes, longitudes):
...
-    # solar_declination_angle returns degrees
-    declination = np.deg2rad(declination)
+    xp = array_namespace(latitudes, longitudes)
+    declination = xp.asarray(declination)
+    latitudes = xp.asarray(latitudes)
+    longitudes = xp.asarray(longitudes)
 
-    latitudes = np.deg2rad(latitudes)
+    # solar_declination_angle returns degrees
+    # TODO: deg2rad() is not part of the array API standard
```
HCookie, line 80: As `deg2rad` is not part of the standard, are we going to reimplement it, or cast to numpy and then cast back?

src/earthkit/meteo/stats/array/quantiles.py
```diff
@@ -42,38 +44,44 @@ def iter_quantiles(
...
         return
 
     if method == "sort":
-        arr = np.asarray(arr)
-        arr.sort(axis=axis)
-        missing = np.isnan(arr).any(axis=axis)
+        arr = xp.asarray(arr)
+        arr = xp.sort(arr, axis=axis)
+        missing = xp.any(xp.isnan(arr), axis=axis)
+        print("missing", missing.shape)
```
HCookie: Should this be a print or a log?

src/earthkit/meteo/wind/array/wind.py
```diff
@@ -269,19 +288,36 @@ def windrose(speed, direction, sectors=16, speed_bins=[], percent=True):
...
-    direction = np.atleast_1d(direction)
-    dir_step = 360.0 / sectors
-    dir_bins = np.linspace(int(-dir_step / 2), int(360 + dir_step / 2), int(360 / dir_step) + 2)
+    xp = array_namespace(speed, direction)
 
-    res = np.histogram2d(speed, direction, bins=[speed_bins, dir_bins], density=False)[0]
+    # TODO: atleast_1d is not part of the array API standard
+    speed = xp.atleast_1d(speed)
+    direction = xp.atleast_1d(direction)
+
```
HCookie: As you've added `histogram2d` as a fallback, would it make sense to also add `atleast_1d`?

tests/thermo/test_thermo.py
```diff
@@ -7,13 +7,804 @@
...
+    p = p * 100
+    mr = thermo.array.mixing_ratio_from_vapour_pressure(vp, p)
+
+    assert array_backend.allclose(mr, v_ref, equal_nan=True)
+
+
+@pytest.mark.parametrize("array_backend", ARRAY_BACKENDS)
+@pytest.mark.parametrize("phase", ["mixed", "water", "ice"])
+def test_saturation_vapour_pressure_1(phase, array_backend):
+    ref_file = "sat_vp.csv"
```
HCookie, line 165: Have these reference files been added?

HCookie (commented):
Looks great.
I understand what your were saying about the tests now...

src/earthkit/meteo/solar/array/solar.py
sandorkertesz: Since it is in also in torch (and also in jax), and array-api-compat allows calling it on a namespace we can delay this decision until other array backends not having this method are considered.

src/earthkit/meteo/stats/array/quantiles.py
sandorkertesz: This should be removed.

src/earthkit/meteo/wind/array/wind.py
sandorkertesz: Good question. Right now this falls into the same category as deg2rad (see my comment above).

tests/thermo/test_thermo.py
sandorkertesz: Yes, they are in `tests/data/`.

src/earthkit/meteo/extreme/array/sot.py
sandorkertesz: I created an empty implementation for the other backends so now we can use it just like this:

```python
err = xp.seterr(divide="ignore", invalid="ignore")
```

But this could be misleading. So, the other solution is:

```python
if array_api_compat.is_numpy_namespace(xp):
    err = xp.seterr(divide="ignore", invalid="ignore")
```

src/earthkit/meteo/stats/array/quantiles.py
```diff
@@ -42,38 +44,43 @@ def iter_quantiles(
...
+
     if isinstance(which, int):
         n = which
-        qs = np.linspace(0.0, 1.0, n + 1)
+        qs = xp.linspace(0.0, 1.0, n + 1)
     else:
-        qs = np.asarray(which)
+        qs = xp.asarray(which)
 
     if method == "numpy_bulk":
```
corentincarton, line 59: Shouldn't we change this api and rename the methods? This doesn't make sense anymore now that we have the Array API. @oiffrig what do you think?

oiffrig: Not sure what you mean. Happy to rename the methods to something that doesn't contain "numpy" (although we may want to keep a fallback. PProc uses the "sort" method, it should be kept (more memory-efficient than numpy)

corentincarton: What I mean is: does it still make sense to call it numpy when the backend may not be numpy anymore? It could be confusing when users try to use the function with copy arrays for instance.

oiffrig: Yes, let's give official names that don't contain numpy (may be worth keeping the numpy names for backwards compatibility, don't need to appear in the docs though)

sandorkertesz:
In order to share the namespace/backend code between `earthkit-data` and `earthkit-meteo` I moved it to `earthkit-utils`, which is the newest earthkit component. It is now a required dependency.

If it is acceptable the code is ready to be merged into develop.

src/earthkit/meteo/score/array/correlation.py
```diff
@@ -6,30 +6,42 @@
...
+    x = xp.asarray(x)
+    y = xp.asarray(y)
     assert x.shape == y.shape
-    x = x - np.mean(x, axis=axis, keepdims=True)
-    y = y - np.mean(y, axis=axis, keepdims=True)
-    x /= np.linalg.norm(x, axis=axis, keepdims=True)
-    y /= np.linalg.norm(y, axis=axis, keepdims=True)
+    x = x - xp.mean(x, axis=axis, keepdims=True)
+    y = y - xp.mean(y, axis=axis, keepdims=True)
+    x /= xp.linalg.norm(x, axis=axis, keepdims=True)
```
oiffrig: Should we change this to `xp.linalg.vector_norm`? I don't see `linalg.norm` in the array API spec

oiffrig (approved):

src/earthkit/meteo/score/array/correlation.py
sandorkertesz: Yes, we should use `xp.linalg.vector_norm`. I'll change it.

PR #23, opened 2024-10-31
Control zero humidity in thermo computations

### LINKED ISSUE #20: Nan values in dewpoint_from_specific_humidity
### What happened?

When the `t_from_es ` function gets called by `dewpoint_from_specific_humidity`, the np.log is called on a division where the denominator can be zero. It raises a silent error and produces nans in the result (es=0 and v is -inf).
If the specific humidity is zero, nans are systematically produced. 

However, shouldn't the dewpoint temperature from a specific humidity of 0.0 be defaulted to some meaningful value, representing the dewpoint temperature of dry air, very low, like 0K? It can be that the specific humidity is very close to 0 and due to rounding, datasets end up with nans, which can reveral rather painful when the data is intended to be used for training of ML models.

```
def t_from_es(self, es):
        v = np.log(es / self.c1)
        return (v * self.c4w - self.c3w * self.t0) / (v - self.c3w)
```

### What are the steps to reproduce the bug?

Call thermo.dewpoint_from_specific_humidity(q=y, p=x) with some 0 values in y and any pressure in Pa.

### Version

v0.6.2

### Platform (OS and architecture)

Linux balfrin-ln002 5.14.21-150400.24.81_12.0.87-cray_shasta_c

### Relevant log output

_No response_

### Accompanying data

_No response_

### Organisation

_No response_

sandorkertesz:
Fixes #20 

This PR adds the `eps` and `out` keyword arguments to 

- dewpoint_from_specific_humidity
- dewpoint_from_relative_humidity
- temperature_from_saturation_vapour_pressure

with the follwoing meaning (e.g. for `dewpoint_from_specific_humidity`):
```rst
def dewpoint_from_specific_humidity(q, p, eps=1e-8, out=None):
...
eps: number
        If ``out`` is not None, return ``out`` when ``q`` < ``eps``.
        If out is None, ``eps`` is ignored and return np.nan for ``q``
        values very close to zero.
out: number or None
        If not None, return ``out`` when ``q`` < ``eps``. If None, ``eps`` is
        ignored and return np.nan for ``q`` values very close to zero.
```

Examples:

```python
>>> from earthkit.meteo.thermo import dewpoint_from_specific_humidity
>>> dewpoint_from_specific_humidity(0., 10000.)
np.nan
>>> dewpoint_from_specific_humidity(0., 10000, out=100)
100
>>> dewpoint_from_specific_humidity(1e-6, 10000, eps=1e-5, out=100)
100
```

### DIFF
--- a/src/earthkit/meteo/thermo/array/thermo.py
+++ b/src/earthkit/meteo/thermo/array/thermo.py
@@ -13,6 +13,10 @@ import numpy as np
 from earthkit.meteo import constants
 
 
+def _valid_number(x):
+    return x is not None and not np.isnan(x)
+
+
 def celsius_to_kelvin(t):
     """Converts temperature values from Celsius to Kelvin.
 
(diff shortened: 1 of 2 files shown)


iainrussell:
Looks like a nice mechanism! I wonder if the kwargs need to be a bit more descriptive, such as `zero_q_eps` and `zero_q_out_value` - what do you think @sandorkertesz?

sandorkertesz:
@iainrussell, I would like to keep it simple. Other, methods already use `eps`, while `out` is coming from numpy methods, e.g. `np.log`.

PR #35, opened 2025-03-10
WIP: Add computations along the vertical

mpvginde:
This PR, adds computations along the vertical.
The computations include:

- [x] Compute pressure at model half- and full-levels from `A` and `B` components and `surface pressure`
- [x] Compute geopotential thickness w.r.t the surface on model full-levels
- [x] Compute the pressure at a certain height level in meters

TODO:
- [ ] Write some tests
- [ ] Fix some formatting

### DIFF
--- a/src/earthkit/meteo/thermo/array/thermo.py
+++ b/src/earthkit/meteo/thermo/array/thermo.py
@@ -1729,3 +1729,22 @@ def wet_bulb_potential_temperature_from_specific_humidity(t, q, p, ept_method="i
         return _EptComp.make(ept_method).compute_wbpt(ept)
     else:
         return temperature_on_moist_adiabat(ept, constants.p0, ept_method=ept_method, t_method=t_method)
+
+def specific_gas_constant(q):
+    """Computes the specific gas constant of moist air
+    (specific content of cloud particles and hydrometeors are neglected).
+
+    Parameters
+    ----------
+    q : number or ndarray
+        Specific humidity (kg/kg)
+
+    Returns
+    -------
+    number or ndarray
+        specific gas constant of moist air
+    """
+
+    R = constants.Rd + (constants.Rv - constants.Rd) * q
+    return R
+
--- a/src/earthkit/meteo/vertical/array/vertical.py
+++ b/src/earthkit/meteo/vertical/array/vertical.py
@@ -16,6 +16,8 @@ from typing import Union
 import numpy as np
 from numpy.typing import NDArray
 
+from earthkit.meteo.thermo import specific_gas_consant
+
 def pressure_at_model_levels(
     A: NDArray[Any], B: NDArray[Any], surface_pressure: Union[float, NDArray[Any]]
 ) -> Tuple[NDArray[Any], NDArray[Any], NDArray[Any], NDArray[Any]]:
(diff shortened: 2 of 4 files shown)


src/earthkit/meteo/thermo/array/thermo.py
```diff
@@ -1729,3 +1729,22 @@ def wet_bulb_potential_temperature_from_specific_humidity(t, q, p, ept_method="i
...
+
+    Returns
+    -------
+    number or ndarray
+        specific gas constant of moist air
+    """
+
+    R = constants.Rd + (constants.Rv - constants.Rd) * q
+    return R
+
```
mpvginde, line 1750: @sandorkertesz I have added this function for calculating the gas constant of moist air here.  
Maybe this function already exist somewhere else?

src/earthkit/meteo/vertical/array/vertical.py
```diff
@@ -5,4 +5,187 @@
...
+    Returns
+    -------
+    ndarray
+        pressure at model full-levels
+    ndarray
+        pressure at model half-levels
+    ndarray
+        delta at full-levels
+    ndarray
+        alpha at full levels
```
mpvginde, line 48: @sandorkertesz, I have now 1 function that calculates all these things in one go since they are so connected. But it might be better to create a Class for this like you did for Saturation pressure calculations?

sandorkertesz:
@mpvginde,  thank you very much for this excellent contribution. You posted a few questions, I will check the code and will come back to you with my answers.

src/earthkit/meteo/thermo/array/thermo.py
sandorkertesz: @mpvginde, this function does not exist yet, but there are the very similar `virtual_temperature()` and `virtual_potential_temperature()` methods, which use the "epsilon" formulation.  Ideally, both of these should call your method. I am quite happy to refactor the `virtual_*` code once this PR is merged.

sandorkertesz:
Hi @mpvginde, the code looks fine and can be merged as it is. Thank you very much again for this excellent work. The next step should be to add tests, which I am happy to do.

As for `pressure_on_model_levels()`, I agree it seems to do too many things in one go. One possible solution is to split it into 3 separate functions to:
    - compute the pressure
    - compute delta
    - compute alpha
As you suggested, this could be implemented using a class behind the methods just like the saturation vapour pressure computations do in the `thermo` module.  Please consider this. Maybe this could be a future PR.

Please also see #36 for the wider scope of vertical interpolations in earthkit-meteo. Your `pressure_at_height()` fits well into it but probably it should be generalised to work for other parameters too (on top of pressure) and allow for the options outlined in #36. Again this could be a future work. Please feel to add your ideas/requirement to #36.

Please let me know what you think of it.

mpvginde:
Hi @sandorkertesz,
Thanks for pointing me at #36, I didn't read it before.

If you want I can do some refactoring (but I would need some time to do it) before we merge, but if you think these functions are already useful as is I'm also fine with merging as is. 

I haven't much experience in writing tests for packages, so if you can show an example I'm also happy to contribute to this.

src/earthkit/meteo/vertical/array/vertical.py
```diff
@@ -5,4 +5,187 @@
...
+
+import numpy as np
+from numpy.typing import NDArray
+
+from earthkit.meteo import constants
+from earthkit.meteo.thermo import specific_gas_consant
+
+def pressure_at_model_levels(
+    A: NDArray[Any], B: NDArray[Any], sp: Union[float, NDArray[Any]]
+) -> Tuple[NDArray[Any], NDArray[Any], NDArray[Any], NDArray[Any]]:
```
mpvginde, line 24: Should A and B default to the ECMWF default vertical grid?

sandorkertesz:
Hi @mpvginde, I think this PR is already very useful so I think it should be merged.  As a next step, I suggest adding an option to choose between "arpege" and "ifs" in the computation of "alpha" (currently the "arpege" option is hardcoded). The implication is that it should be an option for the other 2 methods too since they are based on it.  This would be very useful for testing and it is also important to make this option explicit. 

As for the testing,  please see e.g. how "test_virtual_temperature" as an example. Let me know if you need further help. You are welcome to add/improve any tests.

src/earthkit/meteo/vertical/array/vertical.py
sandorkertesz: Good question. Maybe there should be a separate method that could generate A and B for a given vertical grid type? Then it could be passed to `pressure_at_model_levels`. Another question of course is if "model_level" is the right name here, since this is only related to the ECMWF model level. Not sure what would be the right name though.

PR #41, opened 2025-05-07
handle timezone aware datetimes in solar.py

msgomez06:
I ran into the issue when running AIFS from MARS data - this takes care of the error.

### DIFF
--- a/src/earthkit/meteo/solar/array/solar.py
+++ b/src/earthkit/meteo/solar/array/solar.py
@@ -21,10 +21,10 @@ WARNING: This code is incomplete and not tested. DO NOT USE.
 
 def julian_day(date):
     if date.tzinfo is not None and date.tzinfo.utcoffset(date) is not None:
-            day_start = datetime.datetime(date.year, date.month, date.day, tzinfo=date.tzinfo)
+            year_start = datetime.datetime(date.year, 1, 1, tzinfo=date.tzinfo)
     else:
-            day_start = datetime.datetime(date.year, date.month, date.day)
-    delta = date - day_start
+            year_start = datetime.datetime(date.year, 1, 1)
+    delta = date - year_start
     return delta.days + delta.seconds / 86400.0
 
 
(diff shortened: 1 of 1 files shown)


src/earthkit/meteo/solar/array/solar.py
```diff
@@ -20,7 +20,11 @@
 
 
 def julian_day(date):
-    delta = date - datetime.datetime(date.year, 1, 1)
+    if date.tzinfo is not None and date.tzinfo.utcoffset(date) is not None:
+            day_start = datetime.datetime(date.year, date.month, date.day, tzinfo=date.tzinfo)
```
oiffrig:
```suggestion
            day_start = datetime.datetime(date.year, 1, 1, tzinfo=date.tzinfo)
```

src/earthkit/meteo/solar/array/solar.py
```diff
@@ -20,7 +20,11 @@
...
+    else:
+            day_start = datetime.datetime(date.year, date.month, date.day)
```
oiffrig:
```suggestion
            day_start = datetime.datetime(date.year, 1, 1)
```

oiffrig (changes requested):
Hi, thanks for your PR! The dates got mixed up, please review :)
Also, note that we assume all dates are in UTC and therefore use naive datetime objects, so I would advise some care when using time-zone aware dates

msgomez06:
Yes - sorry about mixing up the julian day code with the local time code.

sandorkertesz:
@msgomez06 , please sort out the formatting problem in pre-commit, then your contribution can be merged.

msgomez06:
Hi - I don't know why the workflow failed last time - but I synced the fork and it seems to be fine? Otherwise, it might be that the date isn't being passed as a date into the function - earthkit data uses a utility to transform it if it isn't (https://github.com/ecmwf/earthkit-data/blob/75f086e2ac7f5f9ccd666bfb6a2caacaccf982c5/src/earthkit/data/utils/dates.py).
Should I implement to_datetime into the utils subdirectory and call it?
--edit--
Ah, it's Black that's complaining. I made the edits straight on the site so the edits weren't linted. I'll reformat in a bit.

PR #45, opened 2025-05-28
implement nan_policy for crps

### LINKED ISSUE #43: score.crps returns zero for nan inputs
### What happened?

```
from earthkit.meteo import score
arr1 = np.full((2,5), np.nan)
arr2 = np.full(5, np.nan)

score.crps(arr1, arr2)
>> array([0., 0., 0., 0., 0.])
```

I would expect
```
>> array([np.nan, np.nan, np.nan, np.nan, np.nan])
```

### What are the steps to reproduce the bug?

-

### Version

-

### Platform (OS and architecture)

-

### Relevant log output

```shell

```

### Accompanying data

_No response_

### Organisation

_No response_

Oisin-M:
Close #43 

Implements a `nan_policy` for the CRPS function mimicing scipy's https://docs.scipy.org/doc/scipy/dev/api-dev/nan_policy.html.

Some caveats
- For the propagate case, this does some unnecessary computations for the values we later mask into nans
- For omit, I implemented as:
```
arr1 = np.ones((10,5)) # shape 2,5
arr2 = np.full(5, 2) # shape 5,
arr1[0][0]=np.nan
score.crps(arr1, arr2, 'omit')
>> array([1., 1., 1., 1.])
```
i.e. if any of the ensemble members are missing, we treat that slice as missing. We could also have thought about trying to compute with just the 9 valid ensemble members for this case, but that may be more confusing

### DIFF
--- a/src/earthkit/meteo/score/array/crps.py
+++ b/src/earthkit/meteo/score/array/crps.py
@@ -19,6 +19,9 @@ def crps(x, y, nan_policy="propagate"):
         Ensemble forecast
     y: array-like (n_points)
         Observation/analysis
+    nan_policy: str
+        Determines how to handle nans.
+        Options are 'raise', 'propagate', or 'omit'.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


oiffrig (approved):

sandorkertesz:
@Oisin-M , please add a test for these new options

sandorkertesz (approved):

PR #47, opened 2025-06-02
pressure_at_height_level fix

sandorkertesz:
This PR implemented the following changes:

- fixed code and made tests work for `pressure_at_height_level_part()` and `relative_geopotential_thickness()`. The results are identical to the similar methods in Metview.
- added the `alpha_top` option to `pressure_at_model_levels()` and `pressure_at_height_level_part()` to  initialise alpha on the topmost half level. The possible values are: "ifs" (default) and "arpege"
- added an extra argument `delta` to `relative_geopotential_thickness()`
- changed the order of the "q", "t" arguments to "t","q" in  `pressure_at_height_level_part()` and `relative_geopotential_thickness()`! This was needed to be in line with the rest of earthkit-meteo and reduce the cognitive load.
- made `relative_geopotential_thickness()` array format agnostic
- updated documentation


sandorkertesz:
@mpvginde , I think I found the main problem:

- relative_geopotential_thickness: is not correctly carrying out the summation:

```python
dphi = np.cumsum(np.flip(alpha * R * T, axis=0), axis=0)
```

This code misses the increment in the half-level geopotential. See (2.22)


- there is another problem related to finding bracketing indices

I will try to fix both these issues.

mpvginde:
Thanks for investigating this @sandorkertesz, I never tested the code for height levels higher then 10m.
This shows the relevance of having good tests.

sandorkertesz:
@mpvginde, thanks for the feedback! I will merge this PR. Please note the changes in the function arguments!

PR #54, opened 2025-06-05
Add bootstrapping

oiffrig:
Adds a simple interface for bootstrapping score computations

I am not completely sure `earthkit.meteo.score` is the best place, but at least it feels reasonable to me. Happy to get other suggestions.

# Array example

```python
import numpy as np
from earthkit.meteo.score import pearson

x = ...
y = ...
bootstrapped = bootstrap(pearson, x, y)
np.quantile(bootstrapped, [0.05, 0.95], axis=0)
```

# Xarray example

```python
import xarray as xr

x = xr.DataArray(..., dims=["number", ...])
bootstrapped = bootstrap((lambda arr: arr.mean("number")), x)
bootstrapped.quantile([0.05, 0.95], "sample")
```


martin-janousek-ecmwf:
Interesting! Would it work for the following use case?
**Compute bootstrapped mean of ROC area**
1. Assume an xarray of contingency tables `ct["lat","lon","number","date"]`, where the contingency table elements are stored under the dimension `number` (for CTs of deterministic forecast the size of this dimension is 4 but for ensemble it's `(ens_size+1)*2` (https://sites.ecmwf.int/docs/vtb/annexes.html#annex-2-extended-contingency-table-structure). 
2. Function `roc_area(ct,dim="number")` returns array `roca["lat","lon","date"]` (https://github.com/ecmwf/vtb/blob/master/vtb/metricss/xmetrics.py#:~:text=roc_area)
3. Then to compute a bootstrapped mean of ROC Area along the `date` dimension one has to
   * create sample of `ct` bootstrapped along the axis `date`
   * for each `ct` sample compute `roca`
   * compute mean of each `roca` along `date`
   * return 5-th and 95-th percentiles of `roca` means

Oisin-M:
@oiffrig this looks very cool!

On this question:
> The interface is so far limited to functions that have two (numpy-like) array inputs and runs bootstrapping on one axis of these arrays. Do we need to extend this?

I think the answer is yes. We have functions where the bootstrapping axis isn't necessarily the same e.g. `score.crps` has as input
- x: array-like (n_ens, n_points)
- y: array-like (n_points)
It would be a reasonable use-case to want to bootstrap this over n_points, but these are on different axes (though of course can hack it with axis=-1 for now).

We also have some functions that have more than 2 parameters. I give the example of `wind.w_from_omega` which has as input
- omega : array-like
- t : array-like
- p : array-like

---

I also had a couple of thoughts/questions I thought I would throw out:

- I was actually expecting to return directly the result for every bootstrapped sample, letting the user decide what they want to do after - which could be anything from a `.mean(axis=0)`, quantiles, confidence interval, binning the samples etc. Are we sure that quantiles makes sense as a default?
- Do we need some sort of nan_policy to decide whether or not we also bootstrap missing values? I think it might be good to handle it also here.
- I think we should raise a warning if the user tries to set `n_samples` larger than the input data to sample from. I'm not aware of a case where this makes sense.

corentincarton:
Thanks @oiffrig, really interesting! I didn't know about bootstrapping so it's always good to learn :)

I'm a bit puzzled about the approach though. From what I see, the bootstrappable function that you use as an example here does not reduce the dataset in the axis direction, i.e.
```
f(x(n), y(n)) -> z(n)
```
Couldn't you then bootstrap the result directly instead of bootstrapping the function? That would make the whole process simpler, and would then work for @Oisin-M and @martin-janousek-ecmwf cases.

In order word, I feel the bootstrappable function should be the statistical function that reduces the array (here the array is the output of the metric function) in one or multiple dimensions. But those bootstrappable functions should not be the metrics that compare the experiment with the observation (CRPS, difference, bias, etc.).

I always visualise these evaluation workflows in three steps:
1. Align experiment with observations (interpolation, slice, etc.), `x(lat, lon), y(n) -> x(n), y(n)`
2. Compute metric between x and y, `x(n), y(n) -> z(n)`
3. Compute statistics over one or multiple dimensions (here 1) `z(n) -> w `

In this case, the bootstrapping is happening at the third level, and should be a type of spatial/temporal statistics. If I understand correctly of course. Please correct me if I'm wrong!

martin-janousek-ecmwf:
I must admit I was initially confused as my experience with bootstrapping was exactly along the lines Corentin very clearly described. But then I thought, oh, maybe the bootstrapping was meant as something even more general, like the `difference()` function from the initial comment, where the bootstrapping the differences of two arrays was meant to by done by reshuffling one or both operands and then taking the difference? Which I found interesting, even though I personally was not aware of a use case for it.

Oisin-M:
Ah yes, I got confused with the first example. I agree actually the difference function should **_not_** be bootstrappable, and neither should the `score.crps` or `wind.w_from_omega` examples I gave. The functions that **_should_** be bootstrappable are `quantiles` and `sig_mask`.

The only edge case I'm not sure about is if a function isn't deterministic, but I don't think we have such a case anyway.

oiffrig:
Yes, I tried to come up with a quick example, and got confused. Do disregard the `difference` example. If you have some examples of bootstrapping I'd be happy to see them, after offline discussion with @corentincarton and @Oisin-M there seems to be quite a range of use cases, and it's definitely worth considering the extent of what we want to support.


@martin-janousek-ecmwf
> Interesting! Would it work for the following use case? **Compute bootstrapped mean of ROC area**

>     3. Then to compute a bootstrapped mean of ROC Area along the `date` dimension one has to
>        
>        * create sample of `ct` bootstrapped along the axis `date`
>        * for each `ct` sample compute `roca`
>        * compute mean of each `roca` along `date`
>        * return 5-th and 95-th percentiles of `roca` means

Not as is, but it looks like something we should be able to support. We would need support for xarray (which looks more and more like the reasonable choice to make anyway), and a looser constraint on number on inputs. Then, the updated bootstrapping would be applied on a function that takes `ct` and computes the mean of `roca`:

```python
@enable_bootstrap(dim="date")
def roca_mean(ct, roca_dim="number", mean_dim="date"):
    return roca(ct, dim=roca_dim).mean(dim=mean_dim)

ct = contingency_table()
bresult = roca_mean.bootstrap(ct)
low, high = bresult.quantiles(0.05)
```

Does that look reasonable to you?

Oisin-M:
Correct me if I'm wrong, but I think we don't want to bootstrap the roca_mean function, but rather just the mean function. It's not necessary/desirable to recalculate ROCA for each bootstrapped time sample, since the ROCA doesn't touch the time axis.

From what I understood, the calculation process would be

### 1. Get CT
```python
ct = read_ct(data)
# dim:size  [lon: Nx, lat: Ny, number: M, date: T]
```

### 2. Calculate ROCA
```python
roca = calculate_roca(ct, dim='number')
# dim:size [lat: Nx, lon: Ny, date: T]
```
**Note:** This would be bootstrappable along dimension number, but we only want to bootstrap along dimension time. Therefore, we should not be recalculating ROCA for each sampled time - it should be done once.

### 3. Bootstrap ROCA output
```python
bootstrapped_roca  = bootstrap_samples(roca, dim='date', sample_size=S, n_bootstraps=B)
# dim:size [lat: Nx, lon: Ny, date: S, sample: B]
```

### 4. Calculate mean per timestep
```python
sample_means = calculate_mean(bootstrapped_roca, dim='date')
# dim:size [lat: Nx, lon: Ny, sample: B]
```

### 5. Calculate 5th and 95 quantiles over the samples
```python
quantiles = calculate_quantiles(sample_means, q=[0.05, 0.95], dim='sample')
# dim:size [lat: Nx, lon: Ny, quantile: 2]
```
**Sidenote:** we could also bootstrap this along the dimension sample because it aggregates along that dimension. In practice, this would be pretty weird - we would essentially be trying to estimate the uncertainty in our estimated uncertainty of the mean.

---

Assuming I haven't misunderstood things, I think in the snippet therefore it should be rather
```python
@enable_bootstrap(dim="date")
def mean(roca_vals, mean_dim="date"):
    return roca_vals.mean(dim=mean_dim)

ct = contingency_table()
roca_vals = roca(ct, dim="number")
bresult = mean.bootstrap(roca_vals)
low, high = bresult.quantiles(0.05)
```
if we want to stick with returning this `BootstrapResult` class. Personally, I would be happier getting directly back the xarray, since the user may want to do some odd calculations afterward. There is also already a `.quantile` method for xr dataarrays.

martin-janousek-ecmwf:
But ROCA is computed from a mean (or, equally, a sum) of CTs over time. We want to collect CT at each station over a period of time and only then to compute ROCA (and other CT-based scores like PSS, ETS etc).

martin-janousek-ecmwf:
Could it then work like
```
@enable_bootstrap(dim="date",bdim="sample")
def mean(ct, mean_dim="date"):
    return ct.mean(dim=mean_dim)

ct_vals = contingency_table()
bresult = mean.bootstrap(ct_vals, )
roca_vals = roca(bresult, dim="number")
low, high = roca_vals.quantiles(0.05,dim="sample")
```

Oisin-M:
Okay I see, thanks for the clarification @martin-janousek-ecmwf. I think then the snippet @oiffrig sent is the correct approach?

> @enable_bootstrap(dim="date")
def roca_mean(ct, roca_dim="number", mean_dim="date"):
    return roca(ct, dim=roca_dim).mean(dim=mean_dim)
ct = contingency_table()
bresult = roca_mean.bootstrap(ct)
low, high = bresult.quantiles(0.05)

martin-janousek-ecmwf:
```
@enable_bootstrap(dim="date")
def roca_mean(ct, roca_dim="number", mean_dim="date"):
    return roca(ct, dim=roca_dim).mean(dim=mean_dim)
```
Sorry, but I still can't see this as correct. This is hinting me we first compute ROCA from station's CT and only then the mean. The order is opposite, first we sum up CTs, and only then we apply `roca()`. Sum and roca are not commutative.

Wouldn't it rather be
```
@enable_bootstrap(dim="date")
def roca_mean(ct, roca_dim="number", mean_dim="date"):
    return roca(ct.mean(dim=mean_dim), dim=roca_dim)
```

Oisin-M:
Thanks for the input @martin-janousek-ecmwf. I agree it's not commutative and the calculation order leads to different interpretations, but I do have some confusion around what we're trying to capture.

Happy to defer to your expertise but doesn't averaging before computing ROCA lose temporal resolution? I would have though we would want to do the ROCA per timestep so that we know the full distribution of ROCA values, and then after we find the mean of that distribution. The other way around i.e computing the ROCA on the mean, would result that we only consider the "mean `ct`" and don't take in other information about the `ct` distribution, which I guess is why we would want to bootstrap.

But couldn't we just get the distribution of the `roca` values per timestep directly and find the mean and quantiles of that?
```python
ct = get_ct()
roca_per_timestep = roca(ct) # distribution of roca values

# just directly find metrics about the distribution
mean = mean(roca_per_timestep, dim='time')
low, high = quantiles(roca_per_timestep, dim='time', qs=[0.05, 0.95])
```

martin-janousek-ecmwf:
Well, I suspect my "expertise" can be easily challenged as I am not a statistician and here I am just hiding behind the fact "we have been always doing it this way". And ROCA was probably poorly chosen as an example as both the ROCA of a single step of a single forecast and the monthly mean ROCA of a single step make sense. A different score, like the ETS computed from a "deterministic" 2x2 contingency table, would have been a better choice to make my point, as ETS is the case when one has first to accumulate CTs over some time period and only then can compute the derived statistics (ETS).

That nonetheless does not change the fact more often than not we do not use an arithmetic average of "daily" scores to get their average value over a period of time (for the same lead time). We use arithmetic mean for bias, but monthly RMSE is not an average of daily RMSEs but sqrt(mean(RMSE**2)), for ACC we don't average ACCs but their Fisher-transformed values. And monthly means of ROCA, ETS, PSS etc are computed from month-accumulated CTs. And the biggest pain, the skill scores. 

Honestly, these are kind of a corner cases, always implemented only later on, but it may be useful to be aware of them as they can break the system.

oiffrig:
Thanks both for your input. I'll try to refine the implementation based on this feedback.

Regarding skill scores, I would be keen on having a look at these use cases anyway, just to see how different they are. It would be great if we could support them out of the box!

oiffrig:
Dear all,

I have now added support for xarray. I have also exposed a few functions, notably `resample` which returns input arrays for bootstrapping, and `bootstrap` which does the bootstrapping. I think this should cover a good range of use cases, but please do point me to other examples if you have specific use cases in mind.

Now that we have the `bootstrap` function, this puts into question the usefulness of the `enable_bootstrap` decorator. Is there any point in keeping it around? I feel like it would be better to just make the call to `bootstrap` explicit.

Also, at the moment `bootstrap` does not return a `BootstrapResult`. I feel like this is now quite superfluous, especially with xarray supporting a range of statistics out of the box anyway.

src/earthkit/meteo/score/array/bootstrap.py
```diff
@@ -0,0 +1,165 @@
...
+    xp = array_namespace(*args)
+    arrays = tuple((xp.asarray(arr), axis) for arr, axis in zip(args, dim))
+    n_inputs = x.shape[dim[0]]
+    assert all(y.shape[axis] == n_inputs for y, axis in arrays), (
+        "Input arrays must have the same size along the sampling dimension"
+    )
+    if n_samples is None:
+        n_samples = n_inputs
+    for _ in range(n_iter):
+        indices = [randrange(n_inputs) for _ in range(n_samples)]
```
Oisin-M: Could use `numpy.random.randint` (or now the new approach is `numpy.random.Generator.integers`) so it's faster + vectorized (and then convert to other array backend if needed)?

src/earthkit/meteo/score/array/bootstrap.py
```diff
@@ -0,0 +1,165 @@
...
+
+    Parameters
+    ----------
+    x, *args: array-like
+        Arrays to sample. Must have the same size along ``dim``
+    dim: int or list of int
+        Sample along this dimension index (either same for all or one per argument)
+    n_iter: int
+        Number of bootstrapping iterations
+    n_samples: int or None
```
Oisin-M, line 38: I wonder if `n_bootstraps` or `n_resamples` might be a clearer name than `n_iter`. What do you think?

src/earthkit/meteo/score/array/bootstrap.py
Oisin-M: If we want to be able to sample with**out** replacement, then we would also need to be able to switch this here somewhere and if we went with `numpy.random.Generator.integers` vectorised approach then it would be mostly a matter of switching it to `numpy.random.Generator.choice` instead. I wonder if worth considering this already now? I can imagine wanting to sample without replacement may come up

Oisin-M (approved):
@oiffrig looks nice! I'm looking forward to using it and I like this implementation returning directly xarray or array. Mostly wanted to raise 2 comments
- might be worth using a vectorised method for generating the random samples
- it would be nice to be able to resample with or without replacement

src/earthkit/meteo/score/array/bootstrap.py
oiffrig: I feel like none of the options (including `n_iter`) is fully satisfactory. I'd be in favour of keeping `n_iter` which feels the least confusing to me.

src/earthkit/meteo/score/array/bootstrap.py
oiffrig: We could, but that means reintroducing a dependency on numpy in that part, not sure how I feel about that. It's definitely not a showstopper though. I'll try to implement it.

oiffrig: We might as well, especially if we do the switch to numpy.random

oiffrig: Here's my suggestion: https://github.com/ecmwf/earthkit-meteo/commit/328f8a793696708c7c47c6c9a5be9c1a010ecdc7

I can merge to this branch if that's satisfactory. The main pain point was getting the tests working again 😆

Oisin-M: I'm happy with that approach! Just to say on the point about avoiding forcing numpy, we can also consider adding an `xp.choice` method to our patched `xp` namespaces - we have numpy and cupy already for free, and torch has `torch.multinomial` which looks like can be a drop-in replacement. I'm happy to open a PR for that if you think it's better

src/earthkit/meteo/score/array/bootstrap.py
Oisin-M: Works for me, let's stick with `n_iter`. I agree none of the options are fully satisfactory

src/earthkit/meteo/score/array/bootstrap.py
oiffrig: Merged. It's a good idea to add `xp.choice` indeed, thanks for suggesting! Can I let you take care of it then?

oiffrig:
@Oisin-M thanks for taking the time to review! I'm happy with the PR, is there anything else worth reviewing? Otherwise I can merge

src/earthkit/meteo/score/array/bootstrap.py
Oisin-M: Sure, happy to do that

Oisin-M:
> @Oisin-M thanks for taking the time to review! I'm happy with the PR, is there anything else worth reviewing? Otherwise I can merge

Happy for you to merge!

src/earthkit/meteo/score/array/bootstrap.py
Oisin-M: https://github.com/ecmwf/earthkit-utils/pull/64

PR #56, opened 2025-07-24
replace arccos with acos as per array api spec

Oisin-M:
### Description
arccos isn't part of the array API spec, so we should be using acos: https://data-apis.org/array-api/2024.12/API_specification/generated/array_api.acos.html. Doesn't matter for torch, numpy, cupy since they implement both, but for example tf.arccos doesn't exist, only tf.acos.

### Contributor Declaration


oiffrig (approved):

sandorkertesz:
@Oisin-M, thanks for fixing this!

Issue #61, opened 2025-10-07
Adopt fully array-api-spec compliant code


Oisin-M:
### Is your feature request related to a problem? Please describe.

There are still a number of places where we have to convert code to be array-api-compliant. This issue is to track progress on this
- [x] use of `xp.fmax` and `xp.fmin` (#62)
- [x] use of `xp.arctan2` (#63)
- [x] use of `xp.deg2rad` (#64, ecmwf/earthkit-utils#27)
- [x] use of eku array backends in testing instead of using array-api-compliant approach (ecmwf/earthkit-utils#16)
- [x] `vertical` module using numpy only

PR #62, opened 2025-10-07
fix/avoid_xp_fmax

Oisin-M:
### Description
`xp.fmax` and `xp.fmin` are not part of the array-api-spec. Part of #61.

### Contributor Declaration


src/earthkit/meteo/extreme/array/efi.py
```diff
@@ -74,7 +73,7 @@ def efi(clim, ens, eps=-0.1):
             defimax = xp.where(mask, -acosdiff[icl] - proddiff[icl], 0.0)
             efi += dEFI
             efimax += defimax
-        efimax = xp.fmax(efimax, xp.asarray(eps))
+        efimax = xp.where(efimax > eps, efimax, eps)
```
corentincarton: @Oisin-M Can you do a quick check of the performance loss with this?

src/earthkit/meteo/extreme/array/sot.py
```diff
@@ -39,24 +39,11 @@ def sot_func(qc_tail, qc, qf, eps=-1e-4, lower_bound=-10, upper_bound=10):
...
-    xp.seterr(**err)
-
-    mask_missing = xp.isnan(sot)
-
-    # upper and lower bounds
-    mask2 = xp.logical_and(xp.logical_not(mask_missing), sot < lower_bound)
-    sot[mask2] = lower_bound
-    mask3 = xp.logical_and(xp.logical_not(mask_missing), sot > upper_bound)
-    sot[mask3] = upper_bound
+    sot[sot < lower_bound] = lower_bound
```
corentincarton, line 45: Does this still work well with NaNs?

src/earthkit/meteo/score/array/crps.py
```diff
@@ -60,16 +60,17 @@ def crps(x, y, nan_policy="propagate"):
     diffxx = x[1:] - x[:-1]  # x[i+1]-x[i], size ens-1
 
     # if i == 0
-    zero = xp.asarray(0)
 
     alpha[0] = 0
-    beta[0] = xp.fmax(diffxy[0], zero)  # x(0)-y
+    beta[0] = xp.where(diffxy[0] > 0, diffxy[0], 0)  # x(0)-y
```
corentincarton: Same here. Check that this is not too slow.

src/earthkit/meteo/extreme/array/sot.py
Oisin-M: This is completely equivalent to the above, but less verbose and more performant. Anything nan will return `False` for `sot < lower_bound`, so
```python
xp.logical_and(xp.logical_not(xp.isnan(sot)), sot < lower_bound)
```
is redundant and can just be `sot < lower_bound`. Same for upper bound case.

src/earthkit/meteo/extreme/array/efi.py
Oisin-M: For array `eps` of size `50_000_000`, old code is 1.38x faster. If this is a concern, we can add `xp.fmax` to our patched array-api implementations in earthkit-utils so that we have no performance loss. `fmax` is not part of the spec, but it is available in numpy, cupy, torch, jax so this is maybe worth doing. Let me know what you think

oiffrig: Could we use `xp.maximum` instead? The behaviour with `NaN`s is different, but I think this works in our favour in this case

Oisin-M: This PR was me ensuring I reproduce the old behaviour but I'm very happy to switch to `xp.maximum` if we're happy with slightly different nan handling!

src/earthkit/meteo/score/array/crps.py
oiffrig: I think we should be able to use `xp.minimum`/`xp.maximum` here (and in the following lines) as well, since we explicitly enforce the `nan_policy` anyway

Oisin-M: I agree, except we have some weird edge cases that we don't explicitly handle for the `nan_policy`. An example is if we have `np.inf` and `-np.inf` values in input, then we can actually get `np.nan` values occuring despite having no nans as input. But I think that's a low concern, and could be address in another issue if needs be since it's not related to changes here

Oisin-M:
Test failures are downstream and unrelated to any changes in this PR

Oisin-M:
Hi @oiffrig & @corentincarton - just looking through some older PRs from last year... are you happy for me to merge this or are there other changes outstanding? Merging is blocked for me

oiffrig (approved):
Looks good to me! If @corentincarton is happy with it I can merge

PR #64, opened 2025-10-07
chore: replace deg2rad with array-api-spec version

Oisin-M:
### Description
`xp.deg2rad` is not part of the array-api-spec. Part of #61 

### Contributor Declaration

### DIFF
--- a/src/earthkit/meteo/solar/array/solar.py
+++ b/src/earthkit/meteo/solar/array/solar.py
@@ -12,6 +12,8 @@ import datetime
 import numpy as np
 from earthkit.utils.array import array_namespace
 
+from earthkit.meteo.constants import radian
+
 DAYS_PER_YEAR = 365.25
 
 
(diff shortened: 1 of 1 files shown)


sandorkertesz:
@Oisin-M thank you for this contribution. I think deg2rad() is a very useful function so ideally it should be added to the patched earthkit-utils array_namespace. However, the complication is that it uses a constant form earthkit-meteo, so earthkit-utils should also have a constants module.

Oisin-M:
@sandorkertesz happy to add it to utils, but in that case I might propose some constants should migrate there also, like radian for example. What do you think?

Oisin-M:
PR now at https://github.com/ecmwf/earthkit-utils/pull/27

We may want to delete the constants radian and degree from ekm to avoid duplication.

sandorkertesz:
Well, this is an interesting question. For simplicity, I would keep the `degree` and `radian` in meteo and sort out later what to do with the constants in general. We also have constants in earthkit-geo ...

Oisin-M:
PR in earthkit-utils to add this functionality is here: https://github.com/ecmwf/earthkit-utils/pull/27, and if it is merged then we should close this PR and **not merge**

Oisin-M:
Addressed in ecmwf/earthkit-utils#27

PR #79, opened 2025-11-25
Add vertical interpolation

cfkanesan:
### Description

High level functions implementing vertical interpolation to pressure, potential temperature and arbitrary fields.

### Contributor Declaration


sandorkertesz:
Hi @cfkanesan, thank you very much for this contribution. These would be the first methods implemented in the high-level earthkit-meteo API! 

Please note there is an ongoing development to add somewhat similar methods to the array level interface: #81.  So it would be important to agree on a naming convention for the names and args/kwargs of all these methods.

These are just my initial thoughts for `interpolate_k2p`.

interpolate_k2p
------------------

```python
def interpolate_k2p(
    field: xr.DataArray,
    mode: Literal["linear_in_p", "linear_in_lnp", "nearest_sfc"],
    p_field: xr.DataArray,
    p_tc_values: Sequence[float],
    p_tc_units: Literal["Pa", "hPa"],
) -> xr.DataArray:
```

earhkit-meteo tends to use long method names and short args and kwargs names. So in the same spirit the name could be `interpolate_from_k_to_pressure_levels`. However, I am not sure  if `k` is descriptive enough   (at least I am not familiar with it).

Maybe `k` could be e.g. `model`?  But unfortunately "model" is ambiguous. There are already some methods in `meteo.vertical` doing computations with ECMWF model levels, e.g. `pressure_at_model_levels`. I think those methods should be renamed having e.g. `eta` instead of  `model` in the name. In that case `k` could be kept.

- `field`: maybe this should be `data` to be more generic?
- `p_field: maybe `p_data`?
- `mode`: maybe this should be `interpolation`? As for the options, I suggest `linear`, 'log` and `nearest` for brevity, assuming these names are descriptive enough. 
- `p_tc_values`: I suggest `p_target` 
- `p_tc_units`: I suggest `p_target_units` 

Also consider assigning a default value to `mode` (linear). Maybe also to `p_tc_units`?

sandorkertesz:
Hi @cfkanesan , as the outcome of some recent discussions we suggest the following naming convention and arg/kwarg ordering for the methods in the PR:

- the input vertical levels should be called `sleve` (since as we understand ICON uses SLEVE coords)
- the kwargs of a method should start with the input data/coords followed by the target related parameters
- we suggest "theta" for potential temperature. However, there are methods in "earthkit.meteo.vertical", which use `potential temperature` in the name and `th` as a kwarg. E.g.: `def temperature_from_potential_temperature(th, p)`. We will need to change here `th` to the more descriptive `theta`. As for the method name, we are not sure  what to do, but we would like to use same naming conventions in `vertical` and `meteo`. This requires a separate discussion.

 The proposal
---------------

```python
# current name: interpolate_monotonic
def interpolate_monotonic(
    data,
    coord,
    target_coord,
    interpolation)
```
 
```python
# current name: interpolate_to_pressure_levels
def interpolate_to_pressure_levels(
  data,
  p,
  target_p,
  target_p_units.
  interpolation
)
```

```python
# current name: interpolate_to_any
def interpolate_sleve_to_coord_levels(
    data,
    h,
    target_data,
    target_coord,
    folding_mode
) 
```

```python
# current name: interpolate_to_theta_levels
def interpolate_sleve_to_theta_levels(
    data,
    theta,
    h,
    target_theta,
    target_theta_units,
    folding_mode,
)
````

**Comment by cfkanesan** [comment]
Hi @sandorkertesz, thanks for the feedback. Perhaps `coord` could be renamed to `target_data` to have consistency between `interpolate_sleve_to_coord_levels` and `interpolate_monotonic`? Otherwise, I agree with the suggestions.

**Comment by sandorkertesz** [comment]
> Hi @sandorkertesz, thanks for the feedback. Perhaps `coord` could be renamed to `target_data` to have consistency between `interpolate_sleve_to_coord_levels` and `interpolate_monotonic`? Otherwise, I agree with the suggestions.

Hi @cfkanesan , I think using `target_data` instead of `coord` in `interpolate_monotonic` would be misleading since `coord` describes the source.

**Comment by cfkanesan** [comment]
From my understanding, both `coord` and `target_data` represent the values of the target coordinate field on the same levels as the input data. Perhaps the ordering of the arguments is confusing, because in `interpolate_sleve_to_theta_levels`, the position of this argument is swapped with `h`.

**Comment by sandorkertesz** [comment]
Hi @cfkanesan, having checked the code,  using `target_data` in my original proposal  in `interpolate_sleve_to_coord_levels` is wrong. I would simply use the term `coord` in both cases:

```python
def interpolate_monotonic(
    data,
    coord,
    target_coord,
    interpolation)

def interpolate_sleve_to_coord_levels(
    data,
    h,
    coord,
    target_coord,
    folding_mode
) 

```

... since `coord` describes the coordinates on the same levels as the source `data`  ( and it is always the same type of coordinate as `target_coord`).

src/earthkit/meteo/vertical/interpolation.py
```diff
@@ -0,0 +1,411 @@
...
+        raise ValueError(
+            "target coordinate value "
+            f"out of range (must be in interval [{th_tc_min}, {th_tc_max}]K)"
+        )
+    tc = TargetCoordinates(
+        type_of_level="theta",
+        values=tc_values.tolist(),
+    )
+
+    return interpolate_sleve_to_coord_levels(data, theta, tc, h, folding_mode)
```
sandorkertesz: Should not it be:
```python
return interpolate_sleve_to_coord_levels(data, h, theta, tc, folding_mode)
```

sandorkertesz:
Hi @cfkanesan, please can I ask you to add xarray as a dependency to `pyproject.toml:optional-dependencies.test` and `tests/downstream-ci-requirements.py`? This is needed to run the CI.

src/earthkit/meteo/vertical/interpolation.py
```diff
@@ -0,0 +1,411 @@
...
+def interpolate_monotonic(
+    data: xr.DataArray,
+    coord: xr.DataArray,
+    target_coord: TargetCoordinates,
+    interpolation: Literal["linear", "log", "nearest"] = "linear",
+) -> xr.DataArray:
+    """Interpolate a field to isolevels of a monotonic target field.
+
+    Example for vertical interpolation to isosurfaces of a target field,
+    which is strictly monotonically decreasing with height.
```
sandorkertesz: Does it have to decrease with height?

src/earthkit/meteo/vertical/interpolation.py
```diff
@@ -0,0 +1,411 @@
...
+
+
+def interpolate_to_pressure_levels(
+    data: xr.DataArray,
+    p: xr.DataArray,
+    target_p: Sequence[float],
+    target_p_units: Literal["Pa", "hPa"] = "Pa",
+    interpolation: Literal["linear", "log", "nearest"] = "linear",
+) -> xr.DataArray:
+    """Interpolate a field from model (k) levels to pressure coordinates.
```
sandorkertesz: This seems to work with any vertical coordinate system

sandorkertesz:
@cfkanesan, thank you for adding `xarray` as a test dependency. The CI is still not running properly. It think `netcdf4` should be added in a similar manner.

sandorkertesz:
@cfkanesan, I can see that the code quality checks are failing. I suggest you do this locally in your env:

pip install pre-commit
pre-commit install

Then on each commit code pre-commit checks would run for you. Alternatively you can run them with: 

pre-commit run --all-files

src/earthkit/meteo/vertical/interpolation.py
cfkanesan: No this is not true anymore, good catch

src/earthkit/meteo/vertical/interpolation.py
cfkanesan: Should I just remove the `from model (k) levels` part?

sandorkertesz: Yes, I think that would be enough.

PR #88, opened 2025-12-12
Projection-based weather regimes

chpolste:
### Description

Adds new submodule `earthkit.meteo.regimes` with classes to define a weather regime classification based on patterns and functions to project anomaly fields onto these patterns to compute a regime index, following the approach of [Michel and Rivière (2011)](https://doi.org/10.1175/2011JAS3635.1).

The implementation aims to cater to users at three levels:

- **High-level**: Users that want to apply a regime classification, i.e., compute a regime index, using a preconfigured instance of `RegimePatterns` from elsewhere. These users only need functions `project` and `standardise`.
- **Medium-level**: Users that create a new regime classification by instantiating one of the provided subclasses of `RegimePatterns` (i.e., the required pattern generation is already implemented).
- **Low-level**: Users that create a new regime classification and need to implement a new pattern generation scheme in a class inheriting from `RegimePatterns`.

Classes to represent fixed patterns (`ConstantRegimePatterns`) and base patterns modulated by a custom scalar function (`ModulatedRegimePatterns`) are provided with this first implementation. An example notebook shows how to use the data of [Grams](https://doi.org/10.5281/zenodo.18154492) to recreate their 7-pattern regimes classification and apply it to test data.

Implementations are at array-level with dictionaries used to label outputs based on their association with individual regimes and xarray level using DataArrays for input and output.

### Features to consider in follow-up work

- Higher-level interface compatible with `FieldList`
- Support for grids other than lat-lon
- Automatic regridding and cropping of input data based on the gridspec in the pattern class with earthkit.regrid
- Automatic generation of area weights based on the gridspec
- Tools to extract regime patterns from data (EOFs and clustering methods are most commonly used)

### Contributor Declaration


src/earthkit/meteo/regimes/array/index.py
```diff
@@ -0,0 +1,65 @@
...
+    -------
+    dict[str, array_like]
+        Results of the projection for each regime.
+    """
+    ndim_field = len(patterns.shape)
+    assert field.shape[-ndim_field:] == patterns.shape
+
+    ps = patterns.patterns(**patterns_kwargs)
+
+    if weights is None:
```
oiffrig: The docstring mentions `weights="uniform"`, which is not handled here

sandorkertesz:
@chpolste, thank you very much for this development. Since it is related to geography I wonder if earthkit-meteo is the right place for it or it should be in earthkit-geo instead? What do you think?

chpolste:
The regimes describe geographical patterns within meteorological fields, so it's a bit of both. For me, the meteorological application of regimes as recurrent weather patterns is the primary concern. And I prefer when the vocabulary used in the package matches that of the domain (which to me is meteorology, not geography). If it were to go into earthkit-geo I think the references to "regimes" in the code would have to go in favour of just (geographical) "patterns".

I can see that the spatial aggregation in the computation of the projections/regime index is different to the "grid point-based" approach that most/all existing earthkit-meteo functions take over the spatial domain, so this would in some sense be a "first" for this package.

src/earthkit/meteo/regimes/array/index.py
```diff
@@ -0,0 +1,68 @@
...
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+
+def project(field, patterns, weights, **patterns_kwargs):
+    """Project onto the given regime patterns.
+
+    Parameters
+    ----------
+    field : array_like
```
oiffrig, line 15: I think it's worth documenting the expected array dimensions (and the relationship between the arguments and return dimensions)

oiffrig (approved):
Looks good to me!

chpolste:
I've now included an xarray-interface, anticipating the upcoming structure to support multiple data representations. No fieldlist support is available yet.

Note that the xarray implementation does not rely on the array implementation in the background but reimplements the `projection` and `regime_index` functions. This allows for more flexibility in the broadcasting of inputs, weights and patterns. The pattern classes are shared between the implementations for all data representations.

There is still optimisation potential for the pattern generation. At the moment, the code is more eager than necessary when evaluating the patterns for chunked input data. I might return to this in the future if necessary.

Issue #94, opened 2026-01-27
Implementation of verification scores


Oisin-M:
### Is your feature request related to a problem? Please describe.

Issue to log/discuss/track progress on adding verification functionality into earthkit

For now we ignore a improvements that should be tackled in follow-up work:
- NaN Handling - Currently we don't document or consolidate how metrics deal with NaN inputs. Both in the actual metric computation and optional aggregation via `agg_dims`.
- Consistent API:
    - Allow dataset **and** dataarrays as inputs where it makes sense. For some metrics, scores e.g. restricts inputs to dataarrays.
    - Metrics that can return multiple components of a metric should do so in a consistent manner (variables vs. new dimension, multiple return values, etc.)

**Functions required for vtb**

_deterministic_

- [x] error (#95)
- [x] mean_error (#95)
- [x] root_mean_squared_error (#95)
- [x] mean_squared_error (#95)
- [x] standard_deviation_of_error (#95)
- [x] correlation (#95)
- [x] cosine similarity (#95)

_ensemble_

- [x] spread (#96)
- [x] crps_from_ensemble (#96)
- [ ] continuous_ignorance
- [x] quantile_score (#96)
- [ ] diagonal
- [ ] rps
- [ ] rank_histogram

_with climatology_

- [x] crps_from_gaussian (#96)
- [x] crps_from_cdf (#96)
- [ ] continuous_ignorance_gaussian
- [ ] diagonal_quadrature
- [ ] brier_gaussian
- [ ] brier_quadrature
- [ ] ignorance_gaussian
- [ ] ignorance_quadrature
- [ ] rps_gaussian
- [ ] rps_quadrature

_events_

- [ ] event_occurrences
- [ ] contingency_table

_contingency table_

- [ ] roc_curve
- [ ] area_under_roc
- [ ] brier	
- [ ] ignorance
- [ ] equitable_threat_score
- [ ] frequency_bias
- [ ] peirce_skill_score
- [ ] symmetric_extremal_dependence_index

_from crps a and b components_

- [ ] crps_potential
- [ ] crps_reliability

Other functions needed
- [ ] kge (#98)

PR #95, opened 2026-01-27
add deterministic verification functions from vtb

Oisin-M:
### Description

Implements experimental versions of the deterministic verification functions in https://github.com/ecmwf/earthkit-meteo/issues/94:

- error
- mean_error
- root_mean_squared_error
- mean_squared_error
- standard_deviation_of_error
- cosine_similarity
- pearson_correlation

These functions are still considered experimental and might change without warning.

There are also a few remaining design decisions that will affect these functions, which will be addressed in follow-up PRs:

* **NaN behavior:** Each metric might deal with NaNs in the input data differently and there is no documentation yet for the user that explains the behavior or provides arguments to influence it.
* **Aggregation:** Most metrics we added do not provide an additional `agg_dims` parameter to further aggregate/reduce computed scores along extra dimensions. We might want to add that if users ask for it.
* **Argument Conventions:** Are `over`, `agg_dims`, etc. good names?
* **Xarray Metadata:** We don't provide explicit control over attributes or dataarray names. This is something we might want to align with other earthkit components.

### Contributor Declaration


src/earthkit/meteo/score/deterministic.py
```diff
@@ -0,0 +1,599 @@
...
+        The forecast xarray.
+    obs : xarray object
+        The observations xarray.
+    agg_method : str, optional
+        The aggregation method to apply over `agg_dim`. Default is None, which means no aggregation.
+    agg_dim : str or list of str, optional
+        The dimension(s) over which to aggregate. Default is None.
+    agg_weights : xarray object, optional
+        Weights to apply during aggregation. Default is None.
+
```
oiffrig, line 179: I see that `is_angular` is not documented (same for other functions below). The name is quite straightforward, but it assumes the angles are in degrees, which may be overlooked if you don't check the scores docs.

src/earthkit/meteo/score/correlation.py
oiffrig, line 1: Is there a plan regarding the functions in `ekm.score.array`? I'm happy to keep them around, possibly marking them as deprecated before removing in a later release

tests/score/test_deterministic.py
```diff
@@ -0,0 +1,809 @@
...
+        coords={
+            "valid_datetime": VALID_DATETIMES,
+            "latitude": LATITUDES,
+            "longitude": LONGITUDES,
+        },
+    )
+
+
+@pytest.fixture
+def fcst():
```
oiffrig: This is used only a few times in the tests, is it really worth a fixture?

tests/score/test_deterministic.py
```diff
@@ -0,0 +1,809 @@
...
+    """Forecast: sequential values 0-17."""
+    return make_dataset(np.arange(18.0).reshape(2, 3, 3))
+
+
+@pytest.fixture
+def obs():
```
oiffrig: Same as `fcst`

tests/score/test_deterministic.py
```diff
@@ -0,0 +1,809 @@
...
+    result_da = error(fcst_ds["2t"], obs_ds["2t"])
+    assert isinstance(result_da, xr.DataArray)
+    assert result_da.shape == (2, 3, 3)
+
+
+@pytest.mark.skipif(NO_SCORES, reason="Scores tests disabled")
+def test_error_with_aggregation():
+    # Error varies by dimension so we can verify correct aggregation axis
+    # Rows (lat): [1,1,1], [2,2,2], [3,3,3] -> mean over lat = [2,2,2]
+    # Cols (lon): [1,2,3], [1,2,3], [1,2,3] -> mean over lon = [2,2,2]
```
oiffrig: This comment is more confusing than anything else, also the mean over lon should be `[1, 2, 3]` not `[2, 2, 2]`

tests/score/test_deterministic.py
```diff
@@ -0,0 +1,809 @@
...
+            [[2, 2, 2], [3, 3, 3], [4, 4, 4]],
+        ],
+        dtype=float,
+    )
+    obs_values = fcst_values - error_values
+
+    fcst = make_dataset(fcst_values)
+    obs = make_dataset(obs_values)
+
+    weights = xr.DataArray(
```
oiffrig, line 226: Those weights do not change the mean error since the only value that has a different weight is equal to the unweighted mean

tests/score/test_deterministic.py
```diff
@@ -0,0 +1,809 @@
...
+
+    # Both timesteps have weighted variance 0.16 -> std = 0.4
+    expected = xr.Dataset(
+        {"2t": (["valid_datetime"], np.array([0.4, 0.4]))},
+        coords={"valid_datetime": VALID_DATETIMES},
+    )
+    xr.testing.assert_allclose(result, expected)
+
+
+@pytest.mark.skipif(NO_SCORES, reason="Scores tests disabled")
```
oiffrig: The `pearson_correlation` function doen't seem to depend on scores (same with weights)

pyproject.toml
```diff
@@ -30,6 +30,8 @@ dependencies = [
   "deprecation",
   "earthkit-utils>=0.0.1",
   "numpy",
+  "scores",
```
oiffrig: Didn't you want to make `scores` an optional dependency?

tests/score/test_deterministic.py
```diff
@@ -0,0 +1,809 @@
...
+        expected_values.append(corr)
+
+    expected = xr.Dataset(
+        {"2t": (["valid_datetime"], np.array(expected_values))},
+        coords={"valid_datetime": VALID_DATETIMES},
+    )
+    xr.testing.assert_allclose(result, expected)
+
+
+@pytest.mark.skipif(NO_SCORES, reason="Scores tests disabled")
```
oiffrig: The `cosine_similarity` function doesn't need scores (same below with weights)

oiffrig (commented):
Looking good! I'm happy to see this grow :)

pyproject.toml
Oisin-M: Good point. We debated a bit but I agree it would make more sense for now at least

src/earthkit/meteo/score/correlation.py
Oisin-M: I think this a bit of an open question but there is an argument for adding array implementations of all functions (I am thinking about ML applications requiring autograd). Of course, much more work...

src/earthkit/meteo/score/deterministic.py
Oisin-M: Good point. I have added the following to the docs for the relevant functions:
```
    is_angular : bool, optional
        Whether the data represents angular quantities in degrees. Default is False.
```

tests/score/test_deterministic.py
Oisin-M: Nice catch - I updated it now to avoid this

tests/score/test_deterministic.py
Oisin-M: Updated the comment with that correction but happy to remove if it's not useful

tests/score/test_deterministic.py
Oisin-M: Removed now

pyproject.toml
```diff
@@ -60,6 +59,7 @@ optional-dependencies.test = [
   "pytest",
   "pytest-cov",
   "requests",
+  "scores",
```
oiffrig: This should be in its own dependency group, rather than in `test`

oiffrig (changes requested):

pyproject.toml
Oisin-M: I've added a new group and matched the previous approach of nesting the dependencies e.g.
```
optional-dependencies.test = [
  "earthkit-meteo[samples,scores]",
  "pytest",
  "pytest-cov",
]
```
but to be honest I have some question marks over how good an idea this is because presumably this pulls from **pypi**

oiffrig (approved):

PR #96, opened 2026-01-30
Feat/xarray ensemble scores

Oisin-M:
### Description

Works towards #94 

- [x] docstrings
- [x] tests
- [x] credit scores
- [ ] (nice to have) document nan behaviour

### Contributor Declaration


tests/score/xarray/test_ensemble.py
```diff
@@ -0,0 +1,591 @@
...
+    )
+
+    result = crps_from_cdf(fcst, obs, over="threshold", return_components=True)
+    total_only = crps_from_cdf(fcst, obs, over="threshold", return_components=False)
+
+    assert set(result.data_vars) == {
+        "total",
+        "underforecast_penalty",
+        "overforecast_penalty",
+    }
```
andreas-grafberger, line 665: TODO: We probably want to align this everywhere, and e.g. always return components via a `component` dim

Oisin-M:
Logging some offline discussions: we plan to unify `crps_from_ensemble` and `crps_from_cdf` to both return `xr.Datasets` and take as input `xr.DataArrays`. That way everything is consistent, and solves the issue that we cannot return a `component` dimension for the Hersbach decomposition

tests/score/xarray/test_ensemble.py
Oisin-M: Have gone with the alternative approach - always return a dataset with variables for each of the components. For the Hersbach decomposition, we have an extra ensemble dimension for the `alpha` and `beta` that the CRPS itself does not have, so this idea of returning components in a `component` dim only is infeasible there

src/earthkit/meteo/score/xarray/ensemble.py
```diff
@@ -0,0 +1,589 @@
...
+            threshold_weight_fill_method="forward",
+            integration_method="exact",
+            reduce_dims=reduce_dim,
+            preserve_dims=None,
+            weights=None,
+            include_components=return_components,
+        )["total"]
+
+
+def ginis_mean_diff(fcst, over):
```
Oisin-M: Do we want to export this at the toplevel or is this an internal function only? It doesn't use observations so it feels a little out of place here to me. Anyhow, if it needs to be exported at the toplevel, then we really ought to add a test. If it's not needed, I would vote to make internal only and then it's an implementation detail for `crps_from_ensemble` that we are checking via the tests for the Hersbach decomposition.

Oisin-M: This is not required. I am therefore leaving it as internal only

src/earthkit/meteo/score/xarray/ensemble.py
```diff
@@ -0,0 +1,561 @@
...
+    - :math:`s_i` is the spread.
+
+    Parameters
+    ----------
+    fcst : xarray object
+        The forecast xarray.
+    over : str or list of str
+        The dimension(s) over which to compute the spread.
+    reference : xarray object, optional
+        The reference xarray to compare against. If not provided, the mean of the forecastover `over` is used.
```
oiffrig:
```suggestion
        The reference xarray to compare against. If not provided, the mean of the forecast over `over` is used.
```

src/earthkit/meteo/score/xarray/ensemble.py
```diff
@@ -0,0 +1,561 @@
...
+        reference = fcst.mean(dim=over)
+    else:
+        if over in reference.dims:
+            reference = reference.squeeze(over)
+    return ((fcst - reference) ** 2).mean(dim=over) ** 0.5
+
+
+def quantile_score(fcst: T, obs: T, tau: float, over: str | list[str]) -> T:
+    r"""
+    Calculates the quantile score of a forecast compared to a observations.
```
oiffrig:
```suggestion
    Calculates the quantile score of a forecast compared to a set of observations.
```

src/earthkit/meteo/score/xarray/ensemble.py
```diff
@@ -0,0 +1,561 @@
...
+        \operatorname{CRPS}\left[ \mathcal{N}(\mu, \sigma^2), o \right] = &\sigma \left\{ \frac{o - \mu}{\sigma} \left[ 2 \Phi \left( \frac{o-\mu}{\sigma} \right) - 1\right] \right. \\
+        &\left. +2\phi\left( \frac{o - \mu}{\sigma}  \right) - \frac{1}{\sqrt{\pi}} \right\}
+        \end{align*}
+
+    where:
+
+    - :math:`\mathcal{N}(\mu, \sigma^2)` is the probabilistic (Gaussian) forecast,
+    - :math:`o` are the observations,
+    - :math:`\phi\left( (o - \mu)/\sigma \right)` denotes the probability density function of the normal distribution with mean 0 and variance 1 evaluated at the normalized prediction error, :math:`(o - \mu)/\sigma`,
+    - :math:`\Phi\left( (o - \mu)/\sigma \right)` denotes the cumulative distribution function of the normal distribution with mean 0 and variance 1 evaluated at the normalized prediction error, :math:`(o - \mu)/\sigma`.
```
oiffrig:
```suggestion
    - :math:`\phi\left( (o - \mu)/\sigma \right)` denotes the probability density function of the normal distribution with mean 0 and variance 1 evaluated at the normalised prediction error, :math:`(o - \mu)/\sigma`,
    - :math:`\Phi\left( (o - \mu)/\sigma \right)` denotes the cumulative distribution function of the normal distribution with mean 0 and variance 1 evaluated at the normalised prediction error, :math:`(o - \mu)/\sigma`.
```

oiffrig (approved):

PR #97, opened 2026-02-03
Feature/hackathon extreme

corentincarton:
### Description

Update of extreme module to support xarray.

### Contributor Declaration

### DIFF
--- a/src/earthkit/meteo/extreme/array/efi.py
+++ b/src/earthkit/meteo/extreme/array/efi.py
@@ -21,6 +21,9 @@ from .utils import (
 def efi(clim, ens, eps=-0.1, clim_axis=0, ens_axis=0):
     """Compute Extreme Forecast Index (EFI).
 
+    The reduction axis (ensemble and quantiles) is configurable by the user,
+    but the other dimensions of clim and ens must be aligned and match.
+
     Parameters
     ----------
     clim: array-like
--- a/src/earthkit/meteo/extreme/array/efi.py
+++ b/src/earthkit/meteo/extreme/array/efi.py
@@ -35,7 +35,7 @@ def efi(clim, ens, eps=-0.1, clim_axis=0, ens_axis=0):
     Returns
     -------
     array-like
-        EFI values with the reduction axes removed.
+        EFI values.
     """
 
     xp = array_namespace(clim, ens)
--- a/src/earthkit/meteo/extreme/array/efi.py
+++ b/src/earthkit/meteo/extreme/array/efi.py
@@ -12,23 +12,30 @@
 
 from earthkit.utils.array import array_namespace
 
+from .utils import flatten_extreme_input
+from .utils import validate_extreme_shapes
 
-def efi(clim, ens, eps=-0.1):
-    """Compute Extreme Forecast Index (EFI)
+
+def efi(clim, ens, eps=-0.1, clim_axis=0, ens_axis=0):
+    """Compute Extreme Forecast Index (EFI).
 
     Parameters
     ----------
-    clim: array-like (nclim, npoints)
-        Sorted per-point climatology
-    ens: array-like (nens, npoints)
-        Ensemble forecast
+    clim: array-like
+        Sorted per-point climatology. The reduction axis (quantiles) is set by ``clim_axis``.
+    ens: array-like
+        Ensemble forecast. The reduction axis (ensemble members) is set by ``ens_axis``.
     eps: (float)
         Epsilon factor for zero values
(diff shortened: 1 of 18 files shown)


src/earthkit/meteo/extreme/array/utils.py
```diff
@@ -0,0 +1,41 @@
...
+def flatten_extreme_input(xp: Any, da: Any, axis: int) -> tuple[Any, tuple[int, ...]]:
+    """Move the reduction axis to position 0 and flatten the remaining dimensions."""
+    da = xp.moveaxis(da, axis, 0)
+    out_shape = da.shape[1:]
+
+    npoints = 1
+    for s in out_shape:
+        npoints *= s
+
+    da = xp.reshape(da, (da.shape[0], npoints))
```
cfkanesan:
```suggestion
    da = xp.reshape(da, (da.shape[0], -1))
```

src/earthkit/meteo/extreme/array/efi.py
```diff
@@ -12,37 +12,55 @@
 
 from earthkit.utils.array import array_namespace
 
+from .utils import flatten_extreme_input, validate_extreme_shapes
 
-def efi(clim, ens, eps=-0.1):
-    """Compute Extreme Forecast Index (EFI)
+
+def efi(clim, ens, eps=-0.1, clim_axis=0, ens_axis=0):
```
cfkanesan, line 21: The clim and ens axis are now configurable but I think that there is still an assumption that the other axes are in the same order for both the clim and ens arrays. Perhaps this assumption should be documented.

src/earthkit/meteo/extreme/array/utils.py
```diff
@@ -0,0 +1,41 @@
...
+    return da, out_shape
+
+
+def validate_extreme_shapes(
```
cfkanesan, line 23: This validation does not cover the case where the dimensions were in a different order before being flattened due to the second component of shape being a product of the sizes in the respective dimensions.

src/earthkit/meteo/extreme/array/sot.py
```diff
@@ -40,46 +42,69 @@ def sot_func(qc_tail, qc, qf, eps=-1e-4, lower_bound=-10, upper_bound=10):
...
+        clim_shape=out_shape,
+        ens_shape=ens_shape,
+        clim_axis=clim_axis,
+        ens_axis=ens_axis,
+    )
     # perc = xp.asarray(perc)
 
-    if not (isinstance(perc, int) or isinstance(perc, xp.int64)) or (perc < 2 or perc > 98):
-        raise Exception("Percentile value should be and Integer between 2 and 98, is {}".format(perc))
+    if not (isinstance(perc, int) or isinstance(perc, xp.int64)) or (
```
cfkanesan:
```suggestion
    if not (isinstance(perc, (int, xp.int64)) or (
```

cfkanesan (approved):
Overall the changes look good, I just have some minor comments. The xarray API could have an additional test where there is an additional dim on either side to check that broadcasting is performed correctly. However, the test coverage is already pretty good IMO.

src/earthkit/meteo/extreme/array/sot.py
Oisin-M: Just noting if of interest to generalise this further - the way to get all [signed int dtypes](https://data-apis.org/array-api/2024.12/API_specification/generated/array_api.info.dtypes.html) is
```
xp.__array_namespace_info__().dtypes(kind="signed integer")
```

src/earthkit/meteo/extreme/array/utils.py
corentincarton: Just pushed a change to do the validation before flattening.

src/earthkit/meteo/extreme/array/efi.py
corentincarton: I added a note about that in the docstrings.

src/earthkit/meteo/extreme/array/sot.py
corentincarton: Thanks, I updated the check.

src/earthkit/meteo/extreme/array/efi.py
cfkanesan: Thanks for addressing the comment. Numpydoc also allows for possible exceptions to be documented in the `Raises` section. I think it can be helpful but the main driver should be consistency.

cfkanesan (approved):

PR #98, opened 2026-02-04
Feat/kge

Oisin-M:
### Description
Work towards #94

TODOs:
- [x] add KGE function with support for both modified and unmodified versions
- [x] docs
- [x] tests

Note: There are two different KGE implementations - 1) original KGE from 2009 paper e.g. see [here](https://thibhlln.github.io/hydroeval/functions/hydroeval.kge.html) and 2) modified KGE from 2012 paper e.g. see [here](https://thibhlln.github.io/hydroeval/functions/hydroeval.kgeprime.html). I have decided to have a `method` kwarg like we do for `crps_from_ensemble` which switches between these two formulations. The parameter alpha is denoted as gamma in the modified version also, so I have renamed it for that case. Later would be good to implement the modified version directly in `scores` instead of doing this manual patchwork here.

The return type logic matches what we do for `crps_from_ensemble` and `crps_from_cdf`. Forecast and obs are dataarrays, and the returned kge is also a dataarray, unless `return_components is True` in which case we return a dataset with a separate variable per component.

### Contributor Declaration


src/earthkit/meteo/score/xarray/deterministic.py
```diff
@@ -629,3 +631,102 @@ def _common_valid_mask(*arrays):
...
+            s_rho, s_alpha, s_beta = scaling_factors
+            kge["kge"] = 1 - (
+                (
+                    (s_rho * (kge["rho"] - 1)) ** 2
+                    + (s_alpha * (kge["alpha"] - 1)) ** 2
+                    + (s_beta * (kge["beta"] - 1)) ** 2
+                )
+                ** 0.5
+            )
+            return kge
```
chpolste: This ignores `return_components`. Delete this return and the ones below will do the work.

src/earthkit/meteo/score/xarray/deterministic.py
```diff
@@ -629,3 +631,102 @@ def _common_valid_mask(*arrays):
...
+    fcst : xarray.DataArray
+        The forecast xarray.
+    obs : xarray.DataArray
+        The observations xarray.
+    over : str or list of str
+        The dimension(s) over which to compute the kge.
+    method : str, optional
+        The method to compute the variability term :math:`\alpha`. Can be either "original" or "modified". Default is "modified".
+    return_components : bool, optional
+        Whether to return the individual components (:math:`\rho`, :math:`\alpha`, :math:`\beta`) along with the KGE value. Default is False.
```
chpolste: Missing documentation for the `scaling_factors` parameter.

src/earthkit/meteo/score/xarray/deterministic.py
```diff
@@ -629,3 +631,102 @@ def _common_valid_mask(*arrays):
...
     result = covar / np.sqrt(fs_var2 * ob_var2)
     return result.where(np.isfinite(result))
+
+
+def kge(
+    fcst: T,
+    obs: T,
+    over: str | list[str],
+    method: str = "modified",
+    scaling_factors: Optional[Union[list[float], np.ndarray]] = None,
```
chpolste: 👍  scores is really strict about the types here (the type signature is correct). It will not accept a tuple of three floats, for example. Would be nice to propose a bit more flexibility here upstream.

chpolste (changes requested):

src/earthkit/meteo/score/xarray/deterministic.py
Oisin-M: Good idea, will aim to propose a PR upstream

oiffrig (approved):

chpolste (approved):

tests/score/xarray/test_deterministic.py
```diff
@@ -782,3 +783,219 @@ def test_cosine_similarity_with_weights():
...
+    )
+
+    xr.testing.assert_allclose(result, expected)
+
+
+@pytest.mark.skipif(NO_SCORES, reason="Scores tests disabled")
+def test_kge_modified_return_components(rng):
+    fcst_values = np.arange(18.0).reshape(2, 3, 3)
+    noise = rng.normal(0, 1, size=(2, 3, 3))
+    obs_values = fcst_values + noise
```
andreas-grafberger, line 852: Since we're already hardcoding the expected KGE values (which I think is good), I'm wondering if we should just hardcode the inputs as well. Although relying on the random seed also works, I think it makes the tests harder to interpret and change.

src/earthkit/meteo/score/xarray/deterministic.py
```diff
@@ -630,3 +632,110 @@ def _common_valid_mask(*arrays):
...
     result = covar / np.sqrt(fs_var2 * ob_var2)
     return result.where(np.isfinite(result))
+
+
+def kge(
+    fcst: T,
+    obs: T,
+    over: str | list[str],
+    method: Literal["original", "modified"] = "modified",
+    scaling_factors: Optional[Union[list[float], np.ndarray]] = None,
```
andreas-grafberger: Can we also add unit tests for the `scaling_factors` parameter (optimally for the original and modified variant)? Just to make sure we have tests for the entire public API.

src/earthkit/meteo/score/xarray/deterministic.py
```diff
@@ -630,3 +632,110 @@ def _common_valid_mask(*arrays):
...
+        The KGE between the forecast and observations.
+    """
+    if method not in ("original", "modified"):
+        raise ValueError("method must be one of 'original' or 'modified'")
+    scores = _import_scores_or_prompt_install()
+
+    if not (method == "modified" or return_components):
+        return scores.continuous.kge(
+            fcst, obs, reduce_dims=over, scaling_factors=scaling_factors, include_components=False
+        )
```
andreas-grafberger: nit: Could we simplify the control flow and just do the following (and remove the `if method == "modified"` below)? Personally, i think this might make the function a bit easier to read, what do you think?
```suggestion
    if method == "original":
        return scores.continuous.kge(
            fcst, obs, reduce_dims=over, scaling_factors=scaling_factors, include_components=return_components
        )
```

src/earthkit/meteo/score/xarray/deterministic.py
```diff
@@ -630,3 +632,110 @@ def _common_valid_mask(*arrays):
     covar = covar - fc_mean * ob_mean
     result = covar / np.sqrt(fs_var2 * ob_var2)
     return result.where(np.isfinite(result))
+
+
+def kge(
+    fcst: T,
+    obs: T,
```
andreas-grafberger:
```suggestion
    fcst: xr.DataArray,
    obs: xr.DataArray,
```

src/earthkit/meteo/score/xarray/deterministic.py
```diff
@@ -630,3 +632,110 @@ def _common_valid_mask(*arrays):
...
+        \alpha = \frac{\sigma_f}{\sigma_o}
+        \gamma = \frac{\alpha}{\beta}
+
+    where:
+        - :math:`\rho`  = Pearson's correlation coefficient between observed and forecast values.
+        - :math:`f` and :math:`o` are forecast and observed values, respectively
+        - :math:`\mu_f` and :math:`\mu_o` are the means of forecast and observed values, respectively
+        - :math:`\sigma_f` and :math:`\sigma_o` are the standard deviations of forecast and observed values, respectively
+        - :math:`s_\rho`, :math:`s_\alpha` and :math:`s_\beta` are the scaling factors for the correlation coefficient :math:`\rho`,
+          the variability term :math:`\alpha` and the bias term :math:`\beta`
```
andreas-grafberger:
```suggestion
          the variability term :math:`\alpha` (or :math:`\gamma`) and the bias term :math:`\beta`
```

src/earthkit/meteo/score/xarray/deterministic.py
```diff
@@ -630,3 +632,110 @@ def _common_valid_mask(*arrays):
     covar = covar - fc_mean * ob_mean
     result = covar / np.sqrt(fs_var2 * ob_var2)
     return result.where(np.isfinite(result))
+
+
+def kge(
+    fcst: T,
+    obs: T,
+    over: str | list[str],
+    method: Literal["original", "modified"] = "modified",
```
andreas-grafberger: nit: would it make sense to drop a reference to the respective papers below in the docstring? We don't do this for all metrics though (yet), so not a blocker from my side.

andreas-grafberger: As discussed offline, maybe `original` is most user-friendly here, just to be more compliant with other implementations. I don't have a super strong opinion here but think that using `original` would be the safe choice that doesn't result in surprised users.

andreas-grafberger (commented):
Looks good to me! I just noted a few optional nitpicks (please feel free to ignore those) and the two questions we discussed (regarding what the better default `method` is, and missing tests for the `scaling_factors` parameter).

src/earthkit/meteo/score/xarray/deterministic.py
Oisin-M: Yes, it's a good point. Not entirely sure best way to cite papers in the docs - depends on the extensions we use for the docs

Oisin-M: Agree, added

src/earthkit/meteo/score/xarray/deterministic.py
Oisin-M: As discussed offline, we might just remove but if we keep we should indeed add some sort of test. I am inclined to remove since I don't think we have a real use-case for it

src/earthkit/meteo/score/xarray/deterministic.py
Oisin-M: If you're okay with it, I think I will leave this until the docathon where we decide more on how to do this across all packages

andreas-grafberger: Sure, sounds good!

tests/score/xarray/test_deterministic.py
Oisin-M: We use this `rng` also for testing `error`, `mean_error` so I propose synchronising between the testing approaches in a separate PR

src/earthkit/meteo/score/xarray/deterministic.py
Oisin-M: https://github.com/nci/scores/pull/1078

PR #102, opened 2026-02-10
Feat/xarray api for ekm stats

Oisin-M:
### Description
Draft PR to log/open some discussions for the work on adding highlevel API to the `ekm.stats` submodule. This submodule is a bit more tricky than others because of the following:

- `array.iter_quantiles` main use is returning an iterator to avoid memory issues. For xarray, this is not very commonly done since the approach is generally to chunk.
- `array.return_period_to_value` and `array.value_to_return_period` take as input a `GumbelDistribution` object, which itself calculates things on the array level. Obviously, it is not possible to just simply `xr.apply_ufunc` this situation.

Current ideas/work
- `stats.nanaverage` is easy. I have just added a simple `xarray` implementation since it's more straightforward than doing an `xr.apply_ufunc` on the array implementation.
- For `stats.iter_quantiles` I have opted as a first pass not to include an xarray implementation at all since it's unclear if it really makes sense. I think the other possible option is to rename this method to `quantiles` and then provide a trivial xarray implementation that just wraps `xr.DataArray.quantiles`.
- For the `GumbelDistribution`, we may need to have one abstract class and then children classes for the `array`, `xarray` implementations. There would then be "dispatch" class that switches based on inputs.

Also, since it's not very clear if we'll keep this as-is, I am proposing to mark all the methods in this submodule as experimental.

### Contributor Declaration


chpolste:
Expanding on the issues with the GumbelDistribution class already mentioned, some thoughts on three approaches:


### 1) Specialised child classes based on common abstract base class

- For every `earthkit.meteo.foo.Bar`, introduce classes `earthkit.meteo.foo.array.Bar`, `earthkit.meteo.foo.xarray.Bar`, etc. that inherit from the top-level `Bar` class.
- Can use classmethods as constructors and dispatch to the subclasses in these.
- Can keep data representation-independent code in the abstract class to avoid some duplication.

**Issues**:

- Does not apply_ufunc well. E.g., just trying to invoke the array-level implementation of the `fit` constructor from the xarray constructor requires an extra wrapper to extract the information from the distribution *object* that the array-level `GumbelDistribution.fit` returns. It get's worse for methods, since they need an instance of the array-level class as their first argument. Can solve this by introducing additional regular functions for the computations, but that seems like a lot of layers for what currently are quite simple computations.
- Data representation is locked-in. The representation at evaluation time must match that used during construction of the distribution.

**Note**: `dispatch` can be made classmethod-compatible by, e.g., adding a new method `invoke`:

```python
def invoke(self, module, func, *args, **kwargs):
    # Handle bound methods (classmethods)
    if func.__self__ is not None:
        module = getattr(module, func.__self__.__name__)
    return getattr(module, func.__name__)(*args, **kwargs)
```

 and calling this new method in the implementations of `dispatch`. To work, the function object must be passed from `dispatch` rather than just the name.


### 2) One class to rule them all

- A single (data)class for the distribution that works for all data representations.
- Could import this class in the `array`, `xarray`, etc. submodules for convenience so they can be used as drop-in replacements for the general one.
- Use dispatched factories or representation-specific classmethods (from_array, from_dataset, ...) to instantiate the class and dispatched functions to access information from the common distribution object.

**Issues**:

- Not clear how to use apply_ufunc. Where do the `output_core_dims` come from if they aren't available on the distribution class? Can work for reduction (where `output_core_dims` is a subset of `input_core_dims`) or when all new `output_core_dims` are not a function of the dimensions of the distribution object. E.g. if a `GumbelDistribution` is fitted to a field of lat and lon (i.e., `mu` and `sigma` are 2D arrays), where do the lat and lon coordinates come from when calling `return_period_to_value` with a return period? They need to be available to form the output DataArray of output values. They can't be copied from the input.
- It's not easy to find a universal representation. Should it have a fixed format (e.g., arrays with optional metadata in additional attributes) or adapt to the data it was constructed with. E.g., when constructing with an xarray object should that just be stored as is (at the cost of lock-in) or is the data unpacked and separated from its metadata (at the cost of complexity)? If no type-agnostic implementation can be found, the universal representation has to be array-level to avoid a hard dependency on xarray, etc.

**Note**: I'm actually taking this approach for the regimes in #88. I expect that many users will receive a preconfigured `Patterns` object and I want them to be free to choose their own data representation regardless of the representation in the provided object. The projection on the regime patterns is a reduction, so the output dims can be inferred from the input dims.


### 3) Rewrite without own classes and embrace duck-typing

- Use the data types of each data representation to store the information on the distributions. Keep the parameters as variables in a Dataset for xarray, and in a FieldList for earthkit.data. Array-level will have to use arrays with returns of multiple objects for the user to handle.
- Functions that do computations (fit, cdf, etc.) can easily dispatch and the array-level functions can be ufunc'ed in a straightforward manner.
- This approach implies more generally that distributions are identified by conventions on structure and naming instead of having dedicated objects.
- Allows users to work with objects they know. They can easily do further operations like saving to disk (e.g., just call to_netcdf on a Dataset) without having to understand our objects first or relying on us to implement all functionality on the classes.

**Issues**:

- Conventions must be carefully designed, so that representations can uniquely be identified.
- Needs a convention for every representation.
- Exposes the internal structure of the distribution to the user, which means we're locked-in long term or lose backward-compatibility on changes to the conventions.
- Attributes on Datasets/DataArrays can be fragile when operated with. This might render some objects unusable if edited as they won't be recognised any longer.
- Awkward for the array representation, where metadata can't be attached and multiple fields can't be grouped. E.g., how to know that two arrays of parameters together form a Gumbel distribution? Users will have to keep track of this. This isn't necessarily unusual for an array-level API, but provides less convenience than a dedicated class.

chpolste:
Went with option 2 (single class that works with all representations) for the implementation.

Breaking changes for the extreme values API compared to before:

- `axis` arg on fit is now called `over`
- `fit_gumbel` replaces the `GumbelDistribution.fit` classmethod
-  changed argument order: `dist` arg now comes after the `value` and `return_period` args as the dispatcher currently only supports dispatching on the first arg (could be changed back if #113 is merged)
- removed the `freq` arg (for now)

Everything marked experimental.

To consider: should we rename the GumbelDistribution to reflect that it's the right-skewed variant? Scipy calls them gumbel_r and gumbel_l to distinguish. We could alternatively also introduce skew="right" or skew="left" args.

chpolste (approved):

Issue #99, opened 2026-02-05
Add high-level API to earthkit-meteo


Oisin-M:
### Is your feature request related to a problem? Please describe.

Currently, we are lacking xarray support for most submodules in earthkit meteo. The goal of this issue is to add xarray for every submodule in earthkit-meteo, and ideally fieldlist when feasible.

### Describe the solution you'd like

Roughly, this means filling out the xarray handling by taking the underlying array implementation, wrapping it with `earthkit.meteo.utils.decorators.xarray_ufunc` along with some autodetection of dims via `get_dim_from_defaults`. Then the top-level call should `dispatch` to the relevant corresponding implementation.

Pull requests should be opened to merge into the branch `feature/proc-hackathon`, make sure to request a review/tag your reviewer etc.

Submodules to address:
- [x] `ekm.extreme` @corentincarton, to be reviewed by @cfkanesan - #97
- [x] `ekm.score` @Oisin-M and @andreas-grafberger, to be reviewed by @oiffrig - #95, #96 
- [x] `ekm.solar` @corentincarton and @ninaburg, to be reviewed by @EddyCMWF - #118
- [x] `ekm.stats` @Oisin-M and @chpolste, to be reviewed by @corentincarton - #102 
- [x] `ekm.thermo` @ninaburg, to be reviewed by @corentincarton - #101, #106
- [x] `ekm.vertical` @sandorkertesz, to be reviewed by @iainrussell 
   - [x] #79 into develop
   - [x] harmonise xarray and array implementations
- [x] `ekm.wind` @sandorkertesz, to be reviewed by @ninaburg - done as part of #103
- [x] `ekm.regimes` @chpolste, to be reviewed by @sandorkertesz - #88 
- [x] packaging - add xarray, earthkit-data scipy and scores as an **optional** dependency @cfkanesan - #114 
- [x] Currently we have an issue with some failing tests. @sandorkertesz investigated and fixed.
- [x] Once the above is merged into develop, then migrate the decorator to earthkit-utils so they can be leveraged elsewhere
- [x] Code review

ninaburg:
HI @Oisin-M , thanks for creating the issue! I cannot edit the description but here the PR for thermo: https://github.com/ecmwf/earthkit-meteo/pull/101 if we want to keep track of the PR. Thanks!

ninaburg:
[...]

PR #112, opened 2026-02-24
Add xarray support to the `solar` module

ninaburg:
### Description
This PR adds xarray support to the `earthkit.meteo.solar` module.
#### Changes
- Added solar.xarray backend implementing all solar functions using xarray_ufunc.
- Added _dispatch_if_xarray helper to ensure correct dispatch when xarray objects are not the first argument (e.g. scalar date with xarray lat/lon).
- Extended array backend to support numpy.datetime64 scalars and arrays for compatibility with xarray.
- Updated type hints and docstrings to include numpy.datetime64.
- Added tests to validate xarray behaviour and ensure consistency with array backend.

Addresses https://github.com/ecmwf/earthkit-meteo/issues/99.

### Contributor Declaration

### DIFF
--- a/src/earthkit/meteo/solar/array/solar.py
+++ b/src/earthkit/meteo/solar/array/solar.py
@@ -15,57 +15,97 @@ from earthkit.utils.array import array_namespace
 DAYS_PER_YEAR = 365.25
 
 
-def julian_day(date):
-    """Day of year as a fractional value.
+def _julian_day_scalar(d: datetime.datetime) -> float:
+    """Compute fractional day-of-year for a single datetime.
 
     Parameters
     ----------
-    date : array-like
-        Input date.
+    d : datetime.datetime
+        Input datetime.
 
     Returns
     -------
-    array-like
-        Number of days since January 1st of the same year, including
-        fractional day component based on the time of day.
-
-    Notes
-    -----
-    This is not the astronomical Julian Day Number. It is the day-of-year
-    (DOY) expressed as a floating-point value.
+    float
+        Day-of-year as a fractional value, where January 1st 00:00 is 0.0.
     """
-    if date.tzinfo is not None and date.tzinfo.utcoffset(date) is not None:
-        year_start = datetime.datetime(date.year, 1, 1, tzinfo=date.tzinfo)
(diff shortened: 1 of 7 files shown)


Oisin-M (changes requested):
Hi @ninaburg, this looks good! One thing to fix - the `xarray_ufunc_kwargs={"vectorize": True}` is playing badly with datetimes and causing some tests to fail. I think we can just remove it and then it should work.

Other small aside - also ran into the same issue with detecting xarray based on first argument being overly limiting - I opened a PR [here](#113) to make it more flexible, and also optionally dispatch to the array-level implementation. Would be good to come up with something to handle both our cases. At the moment, the PR I opened is just checking a single (configurable) argument and not checking all arguments to see if anything is fieldlist/xarray but we could easily change the logic

src/earthkit/meteo/solar/array/solar.py
```diff
@@ -15,20 +15,98 @@
...
+    # numpy datetime64 array
+    if isinstance(date, np.ndarray) and np.issubdtype(date.dtype, np.datetime64):
+        obj = date.astype("datetime64[us]").astype(object)
+        return np.vectorize(_julian_day_scalar, otypes=[float])(obj)
+
+    # python datetime.datetime
+    return _julian_day_scalar(date)
+
+
+def solar_declination_angle(date) -> tuple[float, float]:
```
EddyCMWF, line 78: Do we not want/need to use the array_namespace api approach with this function?

Oisin-M: Would be great, but not sure we can very easily. Quoting from [array-api purpose and scope](https://data-apis.org/array-api/2024.12/purpose_and_scope.html)
> Standardization of these dtypes is out of scope: bfloat16, extended precision floating point, datetime, string, object and void dtypes.
Rationale: these dtypes aren’t uniformly supported, and their inclusion at this point in time could put a significant implementation burden on libraries. It is expected that some of these dtypes - in particular bfloat16 - will be included in a future version of the standard.

ninaburg:
Closed this PR and opened https://github.com/ecmwf/earthkit-meteo/pull/118  under my personal account instead of the meteoswiss org to allow edits from ECMWF side.

PR #114, opened 2026-03-02
Feature/optional deps

cfkanesan:
- [x] check when to deprecate polyfill
- [x] add comment about add_note rather than use runtimeerror

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,7 +55,9 @@ optional-dependencies.samples = [
   "requests",
 ]
 optional-dependencies.scores = [
+  "scipy",
   "scores",
+  "xarray",
 ]
 optional-dependencies.test = [
   "earthkit-data",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -79,6 +79,9 @@ where = [ "src/" ]
 [tool.setuptools_scm]
 version_file = "src/earthkit/meteo/_version.py"
 
+[tool.ruff]
+lint.per-file-ignores."tests/**.py" = [ "E402" ]
+
 [tool.isort]
 profile = "black"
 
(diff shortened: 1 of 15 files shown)


pyproject.toml
```diff
@@ -55,7 +55,9 @@ optional-dependencies.samples = [
   "requests",
 ]
 optional-dependencies.scores = [
+  "scipy",
   "scores",
+  "xarray",
 ]
 optional-dependencies.test = [
   "earthkit-data",
```
corentincarton, line 63: Shouldn't we point to the scores and samples extras here (or to all)?

corentincarton:
@sandorkertesz, we have many different requirements files (pyproject.toml, environment.yml, test/downstream-ci-requirements.txt, environment-unit-test.yml, etc.). Any idea on how to remove the burden on maintaining all these files?

pyproject.toml
cfkanesan: It's not clear to me whether optional-dependencies supports referencing other extras but perhaps this particular extra `test` should be implemented as a dependency-group instead? 
https://packaging.python.org/en/latest/specifications/dependency-groups/#dependency-groups

Oisin-M: Just jumping in - from what I see for `[project.optional-dependencies]`, what happens is it looks directly on PyPI and not locally.  Therefore if we write 
```python
optional-dependencies.test = [
"earthkit-meteo[scores,samples]
]
```
it can be problematic for development because we need to have those groups set up correctly on the last PyPI release...

With `uv`, there is the `[dependency-groups]` concept in which one can link to other groups properly, but I think we don't want to assume/force we always use `uv`. Therefore I think actually we might be better off just duplicating at the moment?

corentincarton (approved):

PR #163, opened 2026-06-04
feat: add cape cin

VeraChristina:
### Description
Add basic functionality for CAPE/CIN

### Contributor Declaration


src/earthkit/meteo/thermo/array/cape_cin.py
```diff
@@ -0,0 +1,850 @@
...
+        p_sfc,
+        t_sfc,
+        q_sfc,
+        zh_sfc,
+        extra_outputs=extra_outputs,
+        vertical_axis=vertical_axis,
+    )
+
+
+surface_cape_cin.__doc__ += _CAPE_CIN_COMMON_DOCSTRING
```
sandorkertesz: This solution does not work with the API docs generator used in earthkit-meteo. The docstrings must be written for each and every method.

With this the usage of "Additional parameters" can be avoided and the docstrings will be clearer.

src/earthkit/meteo/thermo/thermo.py
```diff
@@ -2801,10 +2515,153 @@ def specific_gas_constant(
...
 
     The function returns an object of the same type as the input arguments.
     """
-    dispatched = dispatch(specific_gas_constant, array=True)
+    dispatched = dispatch(specific_gas_constant, fieldlist=False, array=True)
     return dispatched(q)
+
+
+def surface_cape_cin(
+    p: "ArrayLike",
```
sandorkertesz: The cape-cin computations only work with numpy arrays. So instead of ArrayLike NDarray (or equivalent) should be used for the typehints.

src/earthkit/meteo/thermo/array/cape_cin.py
```diff
@@ -0,0 +1,850 @@
...
+    All arrays have the shape of the horizontal dimensions of the input.
+    """
+
+    p: np.ndarray
+    """Pressure (Pa)."""
+    t: np.ndarray
+    """Temperature (K)."""
+    q: np.ndarray
+    """Specific humidity (kg/kg)."""
+    z: np.ndarray
```
sandorkertesz: Consider using `zh` instead of `z`. In some of the methods below `zh` is used for geopotential height.

pyproject.toml
```diff
@@ -24,7 +24,8 @@ classifiers = [
 dependencies = [
   "deprecation",
   "earthkit-utils>=1.0.0rc0",
-  "numpy"
+  "numpy",
+  "scipy"
```
sandorkertesz, line 28: Just a remark. It requires further discussion to see if we want to make scipy a hard dependency, or it should be just an optional one.

src/earthkit/meteo/thermo/array/cape_cin.py
VeraChristina: done!

src/earthkit/meteo/thermo/thermo.py
VeraChristina: done!

src/earthkit/meteo/thermo/array/cape_cin.py
VeraChristina: Good point! I have updated it to `zh_agl` for clarity -- as we have `zh` in the inputs, and I find it confusing for it to sometimes be absolute and sometimes relative to the ground level (under the same name). I think Ivan preferred to have geopotential height above ground level for the parcel path and origin, but I will double check with him.

src/earthkit/meteo/thermo/array/cape_cin.py
```diff
@@ -0,0 +1,923 @@
...
+
+import dataclasses as dc
+
+import numpy as np
+from scipy import interpolate
+
+from earthkit.meteo import thermo
+from earthkit.meteo.constants import constants
+
+C_pl = 4218.0
```
sandorkertesz: I think C_pl should be part of `constants.py`

VeraChristina: Moved! I also renamed it to c_l, which is how it's called in the reference where the related values c_pd and c_pv are taken from.

src/earthkit/meteo/thermo/array/cape_cin.py
VeraChristina: I have checked with Ivan. He prefers to have either `zh_agl` or `p` in the parcel path, preferably `zh_agl` (and not `p` nor `zh`). So I have removed pressure from the parcel path.

sandorkertesz:
It does not seem to work for profile input e.i. when p, t etc are 1D arrays. This is the error I get:

File "~/git/earthkit-meteo/src/earthkit/meteo/thermo/array/cape_cin.py", line 645, in _run_cape_cin
    result = comp._cape_cin(p, zh, t, q, p_sfc, t_sfc, q_sfc, zh_sfc)
  File "~git/earthkit-meteo/src/earthkit/meteo/thermo/array/cape_cin.py", line 474, in _cape_cin
    self._lift_parcel(p_start, t_start, q_start, p, t, q, zh_agl)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "~/git/earthkit-meteo/src/earthkit/meteo/thermo/array/cape_cin.py", line 297, in _lift_parcel
    cond = (p <= p_start[None, :]) & (p <= p_lcl[None, :])
                 ~~~~~~~^^^^^^^^^
IndexError: too many indices for array: array is 0-dimensional, but 1 were indexed
(dev) ?1 cape %

src/earthkit/meteo/thermo/array/cape_cin.py
```diff
@@ -0,0 +1,1001 @@
...
+        vertical_axis = _normalise_vertical_axis(vertical_axis, p.ndim)
+        p = np.swapaxes(p, 0, vertical_axis)
+        t = np.swapaxes(t, 0, vertical_axis)
+        q = np.swapaxes(q, 0, vertical_axis)
+        zh = np.swapaxes(zh, 0, vertical_axis)
+
+    result = comp._cape_cin(p, zh, t, q, p_sfc, t_sfc, q_sfc, zh_sfc)
+
+    cape = result.cape.squeeze(axis=0) if is_profile else result.cape
+    cin = result.cin.squeeze(axis=0) if is_profile else result.cin
```
VeraChristina, line 725: My preference would be to return 0D numpy arrays in the case of profile inputs as that is consistent with the array API, but I can also see why users might expect floats

VeraChristina (commented):
Good spot on the profile inputs! The API allows profile inputs now. It assumes 0D numpy arrays for the surface inputs when p, t etc. are 1D arrays, but also accepts floats. It returns 0D arrays for outputs which only have horizontal dimensions (like cape and cin). I am open to other ways of expanding the API to allow profiles

PR #164, opened 2026-06-10
FieldList and extended Grid support for regimes

chpolste:
### Description

- [x] Add a FieldList implementation for `regimes.project`.
- [x] Use earthkit-geo's `Grid` to replace grid spec polyfill for the patterns grid specification. Fully backwards compatible as Grid constructor accepts the previously used grid spec dictionaries.
- [x] Enable automatic area-based weights generation for regular-ll grids.
- [x] Enable automatic regridding of fields to match the patterns grid (only FieldList implementation for now).
- [x] Update the documentation.

Makes earthkit-geo a hard requirement of the regimes submodule. A new optional dependency group `regimes` is added and earthkit-geo is also added as a requirement of the `test` dependency group.

### Questions

- The downstream-ci currently skips the regridding test due to MIR not being available. Should I make MIR available in the test environment with an additional dependency?
- The regime indices are currently returned as plain arrays for the fieldlist implementation. This means the results work out-of-the-box with the dispatcher for `regime_index`, which most users will call as a follow-up. It's just not very nice in terms of metadata. There would be more information available to provide proper labels and coordinates for the values. The output of the function is tabular data and I'm wondering what the most suitable earthkit-data-aligned container for that is? A pandas DataFrame? And should it be a generator to accommodate streamed fieldlists? I'd be happy to mark the FieldList implementation as experimental until that is sorted out.

### Contributor Declaration

### DIFF
--- a/src/earthkit/meteo/regimes/fieldlist/index.py
+++ b/src/earthkit/meteo/regimes/fieldlist/index.py
@@ -28,8 +28,7 @@ def project(fields, patterns, weights=None, patterns_coords=None, regrid_to_patt
         have the shape of the patterns. If no weights are specified, area-based
         weights are generated from the patterns grid.
     patterns_coords : Mapping[str,str], optional
-        Mapping of field metadata keys to keyword arguments of the pattern
-        generator.
+        Mapping of pattern-generator keyword arguments to field metadata keys.
     regrid_to_patterns : bool
         Allow regridding of input fields to match the patterns grid. Enabled by
         default.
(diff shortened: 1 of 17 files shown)


src/earthkit/meteo/regimes/fieldlist/index.py
```diff
@@ -0,0 +1,61 @@
+# (C) Copyright 2026- ECMWF and individual contributors.
+
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation nor
+# does it submit to any jurisdiction.
+
+import earthkit.data as ekd
+import earthkit.geo as ekg
```
sandorkertesz, line 10: This should only be imported if `regrid_to_patterns=True`. In this way when `regrid_to_patterns=False` earthkit.geo is not required (as far as I can see).

chpolste: With the proposed changes, the `Patterns` class will always contain a `Grid` from earthkit-geo to infer the pattern shape, so earthkit-geo will have already been imported by the time `project` is called.

chpolste:
Thanks @sandorkertesz for having a look at the PR! I think it'll be better if I split the FieldList support and Grid changes into separate contributions. I'll close here and open new PRs for each.

The current placeholder in the code that interprets the grid spec needs to go at some point but I see that it might be worthwhile to revisit the grid requirement altogether. I'll need a bit more time to think this through.

PR #187, opened 2026-07-27
Add lunar module

sandorkertesz:
### Description

The PR adds a new module `lunar` for lunar computations based on the work of @HCookie in earthkit-data (see: https://github.com/ecmwf/earthkit-data/pull/1059).

Currently only the `array` backend is implemented.

The lunar module features the following public methods:

- singular_distance_to_moon
- distance_to_moon
- delta_distance_to_moon

Note: the name and args of `singular_distance_to_moon` have to be reviewed.

These computations require `astropy`, which was added as the `lunar` optional dependency.

### Contributor Declaration

### DIFF
--- a/src/earthkit/meteo/lunar/array/__init__.py
+++ b/src/earthkit/meteo/lunar/array/__init__.py
@@ -1,0 +1,14 @@
+# (C) Copyright 2021 ECMWF.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+#
+
+"""
+Lunar computation functions operating on numpy arrays.
+"""
+
+from .lunar import *  # noqa
--- a/src/earthkit/meteo/lunar/array/lunar.py
+++ b/src/earthkit/meteo/lunar/array/lunar.py
@@ -59,7 +59,7 @@ def _get_body_xyz(body_name: str, time: "astropy.time.Time", xp: ArrayNamespace)
     body = get_body(body_name, time)
     body_itrs = body.transform_to(ITRS(obstime=time))
 
-    xyz = xp.array([
+    xyz = xp.asarray([
         body_itrs.cartesian.x.to(astropy_units.km).value,
         body_itrs.cartesian.y.to(astropy_units.km).value,
         body_itrs.cartesian.z.to(astropy_units.km).value,
@@ -93,7 +93,7 @@ def _get_observer_xyz(
         lon=longitudes * astropy_units.deg, lat=latitudes * astropy_units.deg, height=0 * astropy_units.m
     )
     obs_itrs = loc.get_itrs(obstime=time)
-    obs_xyz = xp.array([
+    obs_xyz = xp.asarray([
         obs_itrs.cartesian.x.to(astropy_units.km).value,
         obs_itrs.cartesian.y.to(astropy_units.km).value,
         obs_itrs.cartesian.z.to(astropy_units.km).value,
--- a/src/earthkit/meteo/lunar/array/lunar.py
+++ b/src/earthkit/meteo/lunar/array/lunar.py
@@ -72,12 +72,12 @@ def _get_body_xyz(body_name: str, time: "astropy.time.Time", xp: ArrayNamespace)
 def _get_observer_xyz(
     time: "astropy.time.Time", latitudes: NDArrayLike, longitudes: NDArrayLike, xp: ArrayNamespace
 ) -> NDArray:
-    """Get the ITRS cartesian coordinates of surface observers on Earth at a given time.
+    """Get the :term:`ITRS` cartesian coordinates of surface observers on Earth at a given time.
 
     Parameters
     ----------
     time : astropy.time.Time
-        The observation time (used to set the ITRS obstime).
+        The observation time (used to set the :term:`ITRS` obstime).
     latitudes : NDArrayLike
         Latitudes of the observer(s) in degrees.
     longitudes : NDArrayLike
@@ -88,7 +88,8 @@ def _get_observer_xyz(
     Returns
     -------
     xyz : NDArrayLike (shape (3, N))
-        ITRS cartesian coordinates of the observer(s) in km.
+        :term:`ITRS` cartesian coordinates of the observer(s) in km.
+
     """
     loc = EarthLocation.from_geodetic(lon=longitudes, lat=latitudes)
 
--- a/src/earthkit/meteo/lunar/array/lunar.py
+++ b/src/earthkit/meteo/lunar/array/lunar.py
@@ -17,8 +17,9 @@ from typing import (
 )
 
 from earthkit.utils.array import array_namespace
+from numpy.typing import NDArray
 
-ArrayLike: TypeAlias = Any
+NDArrayLike: TypeAlias = NDArray | float
 ArrayNamespace: TypeAlias = Any
 
 _ASTROPY_AVAILABLE = True
@@ -69,29 +70,32 @@ def _get_body_xyz(body_name: str, time: "astropy.time.Time", xp: ArrayNamespace)
 
 
 def _get_observer_xyz(
-    time: "astropy.time.Time", latitudes: ArrayLike, longitudes: ArrayLike, xp: ArrayNamespace
-) -> Any:
+    time: "astropy.time.Time", latitudes: NDArrayLike, longitudes: NDArrayLike, xp: ArrayNamespace
+) -> NDArray:
     """Get the ITRS cartesian coordinates of surface observers on Earth at a given time.
 
     Parameters
     ----------
     time : astropy.time.Time
         The observation time (used to set the ITRS obstime).
-    latitudes : array-like
+    latitudes : NDArrayLike
         Latitudes of the observer(s) in degrees.
-    longitudes : array-like
+    longitudes : NDArrayLike
         Longitudes of the observer(s) in degrees.
     xp : ArrayNamespace
         The array namespace (e.g., numpy, cupy).
 
     Returns
     -------
-    xyz : array-like (shape (3, N))
+    xyz : NDArrayLike (shape (3, N))
         ITRS cartesian coordinates of the observer(s) in km.
     """
-    loc = EarthLocation.from_geodetic(
-        lon=longitudes * astropy_units.deg, lat=latitudes * astropy_units.deg, height=0 * astropy_units.m
-    )
+    # loc = EarthLocation.from_geodetic(
+    #     lon=longitudes * astropy_units.deg, lat=latitudes * astropy_units.deg, height=0 * astropy_units.m
+    # )
+
+    loc = EarthLocation.from_geodetic(lon=longitudes, lat=latitudes)
+
     obs_itrs = loc.get_itrs(obstime=time)
     obs_xyz = xp.asarray([
         obs_itrs.cartesian.x.to(astropy_units.km).value,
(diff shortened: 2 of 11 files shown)


src/earthkit/meteo/lunar/array/lunar.py
```diff
@@ -0,0 +1,223 @@
...
+    with no reference to the latitude and longitude of the observer.
+
+    Parameters
+    ----------
+    date : datetime.datetime
+        The date and time for which to compute the distance.
+    latitudes : array-like
+        Latitudes, used only for array namespace and device inference.
+    longitudes : array-like
+        Longitudes, used only for array namespace and device inference.
```
Oisin-M: Not sure about having these as inputs if they're not used. I agree the array namespace detection + device inference is a problem, but I would probably just make this return numpy

src/earthkit/meteo/lunar/array/lunar.py
```diff
@@ -0,0 +1,223 @@
...
+    -------
+    distances : array-like (shape (N,))
+        Distances from observer(s) to the target body in km.
+    """
+    if observer_xyz.ndim == 1:
+        observer_xyz = observer_xyz[:, xp.newaxis]  # shape (3, 1)
+    if target_xyz.ndim == 1:
+        target_xyz = target_xyz[:, xp.newaxis]  # shape (3, 1)
+    diff = observer_xyz - target_xyz  # shape (3, N)
+    distances = xp.asarray(xp.linalg.norm(diff, axis=0), device=device)  # shape (N,)
```
Oisin-M: Wouldn't it make more sense to put `observer_xyz` and `target_xyz` to the correct device and then do the calculations rather than first doing the calculations and then putting on device? Here granted it's trivial but if device was gpu I would expect to run the math on gpu and not cpu here

src/earthkit/meteo/lunar/array/lunar.py
```diff
@@ -0,0 +1,223 @@
...
+        The array namespace (e.g., numpy, cupy).
+
+    Returns
+    -------
+    xyz : array-like (shape (3, N))
+        ITRS cartesian coordinates of the observer(s) in km.
+    """
+    loc = EarthLocation.from_geodetic(
+        lon=longitudes * astropy_units.deg, lat=latitudes * astropy_units.deg, height=0 * astropy_units.m
+    )
```
Oisin-M: Does this actually support arrays other than numpy?

Oisin-M (commented):
Looks good! Put some small comments on how backend/device agnostic we can make the functions.

Also one more general question from me is if we should support lists/tuples etc. in the array implementations? Currently we have code something like
```python
xp = array_namespace(latitudes, longitudes)
latitudes = xp.asarray(latitudes)
longitudes = xp.asarray(longitudes)
device = xp.device(latitudes)
```
which is potentially either not needed if we don't need to supports lists etc., **or** probably a good candidate for having a little decorator or similar in utils to reduce such boilerplate. What are your thoughts?

src/earthkit/meteo/lunar/array/lunar.py
```diff
@@ -0,0 +1,223 @@
...
+        observer_xyz = observer_xyz[:, xp.newaxis]  # shape (3, 1)
+    if target_xyz.ndim == 1:
+        target_xyz = target_xyz[:, xp.newaxis]  # shape (3, 1)
+    diff = observer_xyz - target_xyz  # shape (3, N)
+    distances = xp.asarray(xp.linalg.norm(diff, axis=0), device=device)  # shape (N,)
+
+    return distances
+
+
+def singular_distance_to_moon(date: datetime.datetime, latitudes: ArrayLike, longitudes: ArrayLike) -> Any:
```
Oisin-M: Name isn't that intuitive to me. I am sort of wondering if this shouldn't actually be a separate function but more like a single `distance_to_moon` where latitudes and longitudes are optional and if they are not passed, then you get the current `singular_distance_to_moon` behaviour?

src/earthkit/meteo/lunar/array/lunar.py
sandorkertesz: I agree I would return just numpy here. It is somewhat similar to `vertical/array/hybrid.py:hybrid_level_parameters()`, where we decided not to pass the backend as an arg but return  numpy and let the user convert it to the desired array format. 

Whatever solution we choose it should be recorded somewhere as a future developer's guidance.

src/earthkit/meteo/lunar/array/lunar.py
sandorkertesz: I agree this name should be changed, e.g. to the long but descriptive `distance_from_earth_centre_to_moon(date)`. Your solution of unifying it with `distance_to_moon` is also fine with me.

src/earthkit/meteo/lunar/array/lunar.py
sandorkertesz: Actually, all the moon tests are failing with torch cpu/gpu on my macbook, so the usage of `xp` has to be revised.

sandorkertesz: With the latest commit all of the torch cpu tests are passing. However on the gpu (mps) calling `EarthLocation.from_geodetic(...)` fails with:

 TypeError: unsupported operand type(s) for *: 'Tensor' and 'Unit'

src/earthkit/meteo/lunar/array/lunar.py
HCookie: I'd say `distance_from_earth_centre_to_moon` makes sense, I'd rather not have different functions based on the args

src/earthkit/meteo/lunar/array/__init__.py
```diff
@@ -0,0 +1,14 @@
+# (C) Copyright 2021 ECMWF.
```
HCookie, line 1: What's the standard for copyright year updates throughout earthkit?

src/earthkit/meteo/lunar/array/lunar.py
```diff
@@ -0,0 +1,223 @@
...
+
+    Returns
+    -------
+    distances : array-like (shape (N,))
+        Distances from observer(s) to the target body in km.
+    """
+    if observer_xyz.ndim == 1:
+        observer_xyz = observer_xyz[:, xp.newaxis]  # shape (3, 1)
+    if target_xyz.ndim == 1:
+        target_xyz = target_xyz[:, xp.newaxis]  # shape (3, 1)
```
HCookie, line 126: This might have been overly defensively written, if `_get_observer_xyz` returns a shape of `(3, 1)`, none of this is needed

src/earthkit/meteo/lunar/array/lunar.py
sandorkertesz: I wonder if we need the units in the `EarthLocation.from_geodetic(...)`. 
image

src/earthkit/meteo/lunar/array/__init__.py
sandorkertesz: This will be checked

Oisin-M: To be honest for this I would just do a subsequent  PR adding `reuse` to the precommit and then we automate this away rather than spending much time on it now. Tiago added some info on how to do this in the codex and it worked well when I did it for earthkit-hydro, so can also just copy that template. I had picked 2026 but we can obviously easily change (line in ekh is here: https://github.com/ecmwf/earthkit-hydro/blob/5955e45c895152811edc1e57268f21fcc186bc71/REUSE.toml#L7)

src/earthkit/meteo/lunar/array/lunar.py
Oisin-M: I prefer a single `distance_to_moon` function, but either works. We already have different functions based on the args more or less everywhere in earthkit, but `distance_from_earth_centre_to_moon` is a clear name

src/earthkit/meteo/lunar/array/lunar.py
Oisin-M: I think might be time to put developer guidelines in the toplevel earthkit documentation so it's centralised and then have subpackages link to it. I'm happy to open a PR starting this on the toplevel repo if you agree with the approach

src/earthkit/meteo/lunar/array/lunar.py
sandorkertesz: I changed the name to `distance_from_earth_centre_to_moon` and kept only the `date` arg

sandorkertesz:
I made the following changes:

- changed `singular_distance_to_moon()` to `distance_from_earth_centre_to_moon()` and kept only the date arg. It now returns a (numpy) float
- removed the units from: 

    ```python
    loc = EarthLocation.from_geodetic(
            lon=longitudes * astropy_units.deg, lat=latitudes * astropy_units.deg, 
            height=0 * astropy_units.m
        )
    ```

     and simplified as: 


    ```python
    loc = EarthLocation.from_geodetic(lon=longitudes, lat=latitudes)
    ```

     However, even with this when using torch astropy fails with:  `AttributeError: 'torch.dtype' object has no attribute 'kind'`

- since the code only works with numpy I changed all the typehints/docsrings/tests to indicate it and also simplified the code

src/earthkit/meteo/lunar/array/lunar.py
Oisin-M: https://github.com/ecmwf/earthkit/pull/183

src/earthkit/meteo/lunar/array/lunar.py
```diff
@@ -0,0 +1,215 @@
...
+        Latitudes of the observer(s) in degrees.
+    longitudes : NDArrayLike
+        Longitudes of the observer(s) in degrees.
+    xp : ArrayNamespace
+        The array namespace (e.g., numpy, cupy).
+
+    Returns
+    -------
+    xyz : NDArrayLike (shape (3, N))
+        ITRS cartesian coordinates of the observer(s) in km.
```
Oisin-M: I had to look up the acronym so maybe so we just add the full name? If it's known in the community then also fine to leave

src/earthkit/meteo/lunar/array/lunar.py
```diff
@@ -0,0 +1,215 @@
...
+    """
+    # loc = EarthLocation.from_geodetic(
+    #     lon=longitudes * astropy_units.deg, lat=latitudes * astropy_units.deg, height=0 * astropy_units.m
+    # )
```
Oisin-M:
```suggestion
```

src/earthkit/meteo/lunar/array/lunar.py
Oisin-M: Docstrings says it should, so agree can remove this

src/earthkit/meteo/lunar/array/lunar.py
```diff
@@ -0,0 +1,215 @@
...
+
+    Returns
+    -------
+    distances : NDArray (shape (N,))
+        Distances from observer(s) to the target body in km.
+    """
+    if observer_xyz.ndim == 1:
+        observer_xyz = observer_xyz[:, xp.newaxis]  # shape (3, 1)
+    if target_xyz.ndim == 1:
+        target_xyz = target_xyz[:, xp.newaxis]  # shape (3, 1)
```
Oisin-M, line 126:
```suggestion
```

Oisin-M (approved):

src/earthkit/meteo/lunar/array/lunar.py
sandorkertesz: If I do this some tests fail. So I leave it in and the code can be optimised in a separate PR if needed.

src/earthkit/meteo/lunar/array/lunar.py
sandorkertesz: So do you suggest `NDArray | float` here?

src/earthkit/meteo/lunar/array/lunar.py
sandorkertesz: Done.

src/earthkit/meteo/lunar/array/lunar.py
Oisin-M: Ah no sorry, I meant ITRS. NDArrayLike looks fine to me

src/earthkit/meteo/lunar/array/lunar.py
Oisin-M: Works

src/earthkit/meteo/lunar/array/__init__.py
Oisin-M: Comment above is not quite right, but here's a PR adding spdx stuff: #188

src/earthkit/meteo/lunar/array/lunar.py
sandorkertesz: Ah, sorry. As an improvement I created a glossary and added ITRS to it.

